PR #19, opened 2023-04-06
App client refactor and ATC support

### LINKED ISSUE #16: ApplicationClient.call does not work with txn arguments
When using 1.0.0, calling `.call` on an application client with a method that has transactions as one or more of the arguments. The transactions seemingly never get added to the atomic group thus causing the transaction to fail. 

Edit: I have created a test for this here: https://github.com/algorandfoundation/algokit-utils-ts/compare/main...joe-p:algokit-utils-ts:txn_arg. I will work on debugging.

robdmoore:
Fixes #16 cc @joe-p


docs/code/modules/index.md
```diff
@@ -508,7 +508,7 @@ AlgoNode (mainnet)
```
robdmoore, line 508: This change was hanging about in my local and is unrelated to the rest, but harmless to include

src/transaction.ts
```diff
@@ -103,128 +126,108 @@ export const sendTransaction = async function (
...
-  await algod.sendRawTransaction(signedTransactions).do()
+  const result = await atc.execute(algod, sendParams?.maxRoundsToWaitForConfirmation ?? 5)
 
   Config.getLogger(sendParams?.suppressLog).info(`Group transaction (${groupId}) sent with ${transactionsToSend.length} transactions`)
 
   let confirmations: PendingTransactionResponse[] | undefined = undefined
   if (!sendParams?.skipWaiting) {
     confirmations = await Promise.all(
-      transactionsToSend.map(async (t) => await waitForConfirmation(t.txID(), sendParams?.maxRoundsToWaitForConfirmation ?? 5, algod)),
+      transactionsToSend.map(async (t) => (await algod.pendingTransactionInformation(t.txID()).do()) as PendingTransactionResponse),
```
robdmoore, line 168: Unfortunately, ATC doesn't expose the PendingTransactionInfo so I need to make duplicate calls to algod :(

PR #33, opened 2023-04-28
feat(transfer): make fundingSource optional in ensureFunded

daniel-makerx:


example.env
```diff
@@ -0,0 +1,11 @@
+ALGOD_TOKEN=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+ALGOD_SERVER=http://localhost
+ALGOD_PORT=4001
+
+KMD_TOKEN=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
robdmoore, line 5: I don't think we need these three? Or at least we just need KMD_PORT.

Also this file should be called `.env.template` for consistency with `algokit bootstrap` imo

src/transfer.ts
```diff
@@ -50,16 +50,17 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2):
   const currentSpendingBalance = microAlgos(balance - minimumBalanceRequirement.microAlgos)
 
   if (minSpendingBalance > currentSpendingBalance) {
+    const from = fundingSource ?? (await getDispenserAccount(algod))
```
robdmoore, line 53: Should we propagate an optional kmd instance to here so it's possible to control the kmd instance that's used? All of the other things that use kmd have that pattern.

tests/setup.ts
```diff
@@ -1,6 +1,18 @@
+import dotenv from 'dotenv'
+import fs from 'fs'
+import path from 'path'
 import { Config } from '../src/'
 import { consoleLogger } from '../src/types/logging'
 
+let envPath = path.resolve(path.join(__dirname, '..', '.env'))
```
robdmoore, line 7: We can do this in the `scripts` in package.json instead if we want, although not the switch from example env, but if we encourage use of algokit bootstrap all and the right .env.template name then it would work?

PR #37, opened 2023-05-04
feat(app-client): Refactoring types used by ApplicationClient to make them easier to extend

robdmoore:


src/indexer-lookup.spec.ts
```diff
@@ -79,6 +79,7 @@ describe('indexer-lookup', () => {
     const app1 = await algokit.createApp(createParams, algod)
     const app2 = await algokit.createApp(createParams, algod)
     await algokit.createApp({ ...createParams, from: secondAccount }, algod)
+    await waitForIndexer()
```
robdmoore, line 82: Noticed this test broke intermittently, was missing this

src/types/app-client.ts
```diff
@@ -123,12 +139,18 @@ export type AppClientCallArgs =
   | {
       /** If calling an ABI method then either the name of the method, or the ABI signature */
       method: string
-      /** Either the ABI arguments or an object with the ABI arguments and other parameters like boxes */
+      /** A object with the ABI arguments and other parameters like boxes */
```
daniel-makerx, line 142:
```suggestion
      /** An object with the ABI arguments and other parameters like boxes */
```

PR #61, opened 2023-05-26
Use models from sdk

tristanmenzel:
Replaces hand crafted types for objects returned from algod with official models imported from the sdk


src/app.ts
```diff
@@ -230,7 +231,7 @@ export async function updateApp(
     }
   } else {
     const transaction = algosdk.makeApplicationUpdateTxnFromObject({
-      appIndex: appId,
+      appIndex: appId as number,
```
robdmoore, line 234: @tristanmenzel should this be `toNum` too?

src/util.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Converts a value which might be a number or a bigint into a number to be used with apis that don't support bigint.
+ *
+ * Throws an UnsafeConversionError if the conversion would result in an unsafe integer for the Number type
+ * @param value
+ */
+export const toNum = (value: number | bigint) => {
```
robdmoore, line 7: @tristanmenzel should this be `toNumber` instead?

src/types/algod.ts
```diff
@@ -1,145 +1,20 @@
...
-/** Specifies maximums on the number of each type that may be stored https://developer.algorand.org/docs/rest-apis/algod/v2/#applicationstateschema */
-export interface ApplicationStateSchema {
-  /** [nbs] num of byte slices */
-  'num-byte-slice': number
-  /** [nui] num of uints */
-  'num-uint': number
-}
+import { modelsv2 } from 'algosdk'
+
+export const PendingTransactionResponse = modelsv2.PendingTransactionResponse
```
robdmoore, line 3: @tristanmenzel worth keeping the jsdoc comment with the link to the documentation or should we even remove this file altogether? Maybe exposing these types is still handy because the location in the orginal is a little odd...

tristanmenzel: we could probably drop the file all together and just replace usages with the namespaced version

src/util.ts
tristanmenzel: good catch

src/app.ts
tristanmenzel: good catch

PR #80, opened 2023-06-22
feat: Deprecate functions that uses environment configuration by default

inaie-makerx:


src/network-client.ts
```diff
@@ -49,6 +76,21 @@ export function getAlgoNodeConfig(network: 'testnet' | 'mainnet', config: 'algod
...
+ *
+ * @param accountName account name
+ */
+export function getAccountConfig(accountName: string): AccountConfig {
+  if (!process || !process.env) {
+    throw new Error('Attempt to get account with private key from a non Node.js context; not supported!')
+  }
+
+  return {
+    envKey: `${accountName.toUpperCase()}_MNEMONIC`,
```
daniel-makerx, line 89: should this be returning the values of these environment variables?

src/types/network-client.ts
```diff
@@ -9,3 +9,19 @@ export interface AlgoClientConfig {
...
+  algodConfig: AlgoClientConfig
+  /** Indexer client configuration */
+  indexerConfig: AlgoClientConfig
+  /** Kmd configuration */
+  kmdConfig: AlgoClientConfig
+}
+
+/** Config for an account config */
+export interface AccountConfig {
+  envKey: string
```
daniel-makerx, line 25: `accountMnemonic` and `senderMnemonic`?

package.json
```diff
@@ -31,6 +31,7 @@
   "dependencies": {
     "algosdk": "^2.3.0",
     "algosdk2-2": "npm:algosdk@2.2.0",
+    "better-npm-audit": "^3.7.3",
```
daniel-makerx, line 34: this should be a dev dependency?

src/network-client.ts
```diff
@@ -49,6 +76,21 @@ export function getAlgoNodeConfig(network: 'testnet' | 'mainnet', config: 'algod
   }
 }
 
+/** Returns the Account configuration
+ *
+ * @param accountName account name
+ */
+export function getAccountConfig(accountName: string): AccountConfig {
```
daniel-makerx, line 83: Also wonder if this should be called `getAccountConfigFromEnvironment` just so it is clear

daniel-makerx (approved):

PR #83, opened 2023-06-27
fix: changed the audit command on the release pipeline

inaie-makerx:


aorumbayev:
@daniel-makerx @tristanmenzel adding you for quick sanity check as well as i haven't had a lot of work done on these utils repos yet.

aorumbayev (approved):
The change itself is ok but i think commitlint is arguing on the pipeline, the commit messages should all be prefixed in lowercase according to conventions in https://www.conventionalcommits.org/en/v1.0.0/. 

If you can fix the commit names that should fix the error on ci pipeline but otherwise lgtm. Please await for at least one more approval from Tristan or Daniel just to be on a safer side

daniel-makerx (approved):

PR #106, opened 2023-08-20
fix: Added cross-fetch as an explicit dependency to not rely on it coming from algosdk

robdmoore:
This fixes a bug when deploying code to e.g. AWS Lambda and the node_modules is optimised such that cross-fetch isn't able to be imported

Also updated packages to latest versions.


neilcampbell (approved):

aorumbayev (approved):

aorumbayev:
@robdmoore shall we merge this?

PR #107, opened 2023-08-21
chore: adding peer dependency with algosdk

negar-abbasi:


docs/README.md
```diff
@@ -24,6 +24,22 @@ This library can be installed from NPM using your favourite npm client, e.g.:
...
+```
+npx check-peer-dependencies --install
+```
+
+- Alternatively, simply use:
+
+```
+npm install algosdk
+```
+
```
neilcampbell, line 42:
```suggestion
```
```
```
```
```

I think simply saying "Ensure you have it installed in your project" is enough here.

neilcampbell (approved):

robdmoore:
I think this is probaby a breaking change? Should we add that in so major version gets bumped?

robdmoore:
Also, let's pin it to the latest algosdk - `2.5.0`

neilcampbell:
@robdmoore if we pin to that specific version, we will be forcing consumers to use that exact version in their projects. Assuming you're simply suggesting we update the package.json to `"algosdk": "^2.5.0"`

robdmoore:
Yep! `^2.5.0`

PR #111, opened 2023-08-24
fix: splitting compile into separate esm and cjs dists

aorumbayev:
## Proposed changes

- Fixing a weird edge case in https://github.com/algorand/js-algorand-sdk/issues/740 caused by issues with vite module resolution during dev mode without a need for changes in `instanceof` on `js-algorand-sdk` side. An implementation of a fix proposed by @PhearZero


tsconfig.esm.json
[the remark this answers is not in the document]
aorumbayev: Done

tsconfig.cjs.json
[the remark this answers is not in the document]
aorumbayev: Done, i didnt put declaration: false explicitly though as it produces the following:
```
tsconfig.cjs.json:8:5 - error TS5069: Option 'declarationMap' cannot be specified without specifying option 'declaration' or option 'composite'.

8     "declaration": false
```

seems like its false by default anyway if its not provided. Feel free to unresolve though if you think this still needs to be explicitly set to false

aorumbayev:
@PhearZero thanks for PR review! feel free to check how your comments were addressed, 

@tristanmenzel i tested it against generator ts and it works fine but i had to tweak `build:3-copy-pkg-json` to put paths to dedicated folders for main and types records, also needed to explicitly include exports section otherwise importing from testing folder wouldn't work

package.json
```diff
@@ -8,16 +8,39 @@
...
+  },
   "scripts": {
     "build": "run-s build:*",
     "build:0-clean": "rimraf dist coverage",
-    "build:1-compile": "tsc -p tsconfig.build.json",
-    "build:2-copy-pkg-json": "npx --yes @makerx/ts-toolkit@latest copy-package-json",
-    "build:3-copy-readme": "copyfiles README.md dist",
+    "build:1-compile-cjs": "tsc -p tsconfig.cjs.json",
+    "build:2-compile-esm": "tsc -p tsconfig.esm.json",
+    "build:3-copy-pkg-json": "npx --yes @makerx/ts-toolkit@latest copy-package-json --main ./dist/cjs/index.js --types ./dist/types/index.d.ts --custom-sections exports",
```
PhearZero, line 42: ### The generated `package.json` is malformed

- pointing to `./dist` for `main` and `types` top level keys.
- missing top level key `module`
- missing top level key `typesVersions`

Resolutions: Update generator or prefer top level `package.json` with `.npmignore`. The ladder would preserve `npm link` and `workspaces` functionality.

package.json
```diff
@@ -8,16 +8,39 @@
   "engines": {
     "node": ">=16.0"
   },
-  "main": "./index.js",
+  "type": "module",
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
+  "types": "./dist/types/index.d.ts",
+  "typesVersions": {
```
PhearZero, line 15: ### typesVersions

The `typesVersions` needs to be more explicit in the current structure. Types are not resolving
with the wildcard in my test bed. Short term resolutions:

#### Explicit `typesVersions`

```json
{
  "typesVersions": {
    "*": {
      ".": [
        "types/index.d.ts"
      ],
      "types/app": [
        "types/types/app.d.ts"
      ],
      "types/app-client": [
        "types/types/app-client.d.ts"
      ],
      "types/app-spec": [
        "types/types/app-spec.d.ts"
      ],
      "types/transaction": [
        "types/types/transaction.d.ts"
      ]
    }
  }
}
```

Long term strategy would be adding a `./src/types/index.ts` entrypoint and treating `types` as a module (similar to `./src/testing`). It would require deduplication of `StateSpec` and refactoring. Maybe something to consider for `v2`

PhearZero (changes requested):
Working example of what `./dist/package.json` should look like.

```json
{
  "main": "./cjs/index.js",
  "module": "./esm/index.js",
  "types": "./types/index.d.ts",
  "name": "@algorandfoundation/algokit-utils",
  "version": "0.1.0",
  "private": false,
  "description": "A set of core Algorand utilities written in TypeScript and released via npm that make it easier to build solutions on Algorand.",
  "author": "Algorand Foundation",
  "typesVersions": {
    "*": {
      ".": [
        "types/index.d.ts"
      ],
      "types/app": [
        "types/types/app.d.ts"
      ],
      "types/app-client": [
        "types/types/app-client.d.ts"
      ],
      "types/app-spec": [
        "types/types/app-spec.d.ts"
      ],
      "types/transaction": [
        "types/types/transaction.d.ts"
      ]
    }
  },
  "license": "MIT",
  "engines": {
    "node": ">=16.0"
  },
  "files": [
    "**/*"
  ],
  "exports": {
    ".": {
      "import": "./esm/index.js",
      "require": "./cjs/index.js",
      "types": "./types/index.d.ts"
    },
    "./testing": {
      "import": "./esm/testing/index.js",
      "require": "./cjs/testing/index.js",
      "types": "./types/testing/index.d.ts"
    },
    "./index.d.ts": "./types/index.d.ts",
    "./package.json": "./package.json"
  },
  "overrides": {
    "semver": "7.5.2"
  },
  "dependencies": {
    "algosdk": "^2.3.0",
    "buffer": "^6.0.3"
  }
}


```

package.json
aorumbayev: @PhearZero good catch, should be resolved now on the latest commit, retested on generator ts repo

package.json
aorumbayev: Should be explicitly defined now

PhearZero (approved):
LGTM 🕺🏻

aorumbayev:
thanks @PhearZero will await for @tristanmenzel to take a peek and will merge this early next week 👍

tristanmenzel (approved):

tristanmenzel:
LGTM @aorumbayev! :shipit:

PR #133, opened 2023-09-07
fix: adding simulation on failed transactions

negar-abbasi:
After a failed transaction, using simulate to extract more information about what went wrong on a debug mode.

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -1,6 +1,7 @@
 import algosdk, {
   Algodv2,
   AtomicTransactionComposer,
+  EncodedSignedTransaction,
   modelsv2,
   SuggestedParams,
   Transaction,
@@ -289,18 +290,17 @@ export async function performAtomicTransactionComposerDryrun(atc: AtomicTransact
  * @returns The simulation result, which includes various details about how the transactions would be processed.
  */
 export async function performAtomicTransactionComposerSimulate(atc: AtomicTransactionComposer, algod: Algodv2) {
-  const signedTransactions = await atc.gatherSignatures()
-  const decodedSignedTransactions = signedTransactions.map((t) => algosdk.decodeObj(t) as algosdk.EncodedSignedTransaction)
+  const unsignedTransactionsSigners = atc.buildGroup()
+  const decodedSignedTransactions = unsignedTransactionsSigners.map((ts) => algosdk.encodeUnsignedSimulateTransaction(ts.txn))
 
   const simulateRequest = new modelsv2.SimulateRequest({
+    allowEmptySignatures: true,
     txnGroups: [
       new modelsv2.SimulateRequestTransactionGroup({
-        txns: decodedSignedTransactions,
+        txns: decodedSignedTransactions.map((txn) => algosdk.decodeObj(txn)) as EncodedSignedTransaction[],
       }),
     ],
   })
-
-  // Perform simulation
   const simulateResult = await algod.simulateTransactions(simulateRequest).do()
   return simulateResult
 }
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -256,7 +256,7 @@ export const sendAtomicTransactionComposer = async function (atcSend: AtomicTran
       if (simulate.txnGroups[0].failedAt) {
         for (const txn of simulate.txnGroups[0].txnResults) {
           e.traces.push({
-            trace: txn.execTrace,
+            trace: txn.execTrace?.get_obj_for_encoding(),
             appBudget: txn.appBudgetConsumed,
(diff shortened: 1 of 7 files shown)


src/transaction.ts
```diff
@@ -283,6 +282,29 @@ export async function performAtomicTransactionComposerDryrun(atc: AtomicTransact
...
 }
 
+/**
+ * Performs a simulation of the transactions loaded into the given AtomicTransactionComposer.
+ * @param atc The AtomicTransactionComposer with transaction(s) loaded.
+ * @param algod An Algod client to perform the simulation.
+ * @returns The simulation result, which includes various details about how the transactions would be processed.
+ */
+export async function performAtomicTransactionComposerSimulate(atc: AtomicTransactionComposer, algod: Algodv2) {
+  const signedTransactions = await atc.gatherSignatures()
```
robdmoore, line 292: Do we need them to be signed transactions for simulate (I'm not sure)?

src/transaction.ts
```diff
@@ -245,25 +245,24 @@ export const sendAtomicTransactionComposer = async function (atcSend: AtomicTran
...
-        if (txn.appCallRejected()) {
+      const simulate = await performAtomicTransactionComposerSimulate(atc, algod)
+      if (simulate.txnGroups[0].failedAt) {
+        for (const txn of simulate.txnGroups[0].txnResults) {
           e.traces.push({
-            trace: txn.appTrace(),
-            cost: txn.cost,
-            logs: txn.logs,
-            messages: txn.appCallMessages,
+            trace: txn.execTrace,
```
robdmoore, line 258: Previously, we had to use dry run because simulate didn't have traces, I assume the fact we have these fields means that it's up to parity now?

src/transaction.ts
```diff
@@ -245,25 +245,24 @@ export const sendAtomicTransactionComposer = async function (atcSend: AtomicTran
...
+            cost: txn.logicSigBudgetConsumed,
```
robdmoore, line 259: What other fields are there? I suspect we don't want logic sig budget, is there other budget or similar fields we can use?

src/transaction.ts
aorumbayev: @robdmoore based on docstrings/code and documentation on js sdk it seems like they both represent similar notion of traces when txns are app calls. 

See sample ref on simulate execTrace here https://algorand.github.io/js-algorand-sdk/classes/modelsv2.SimulateTransactionResult.html#execTrace

And here is the dryrun definition https://github.com/algorand/js-algorand-sdk/blob/d5ab01407adeba60bf93e315627b2db298de011e/src/dryrun.ts#L335 which seems to be rather similar (i see refs to app calls and logic sig calls)

I am more familiar with simulate though rather than dry run, if you think there is still discrepancy between too we might have to keep dry run too and try to merge them together. But im pretty sure these fields are on par

robdmoore: Yeah I think they are finally at parity on tracing, which is great news!

src/transaction.ts
aorumbayev: @robdmoore there is no direct cost param in simulate its a bit more granular there is appBudgetConsumed and logicSigBudgetConsumed. We can either do a sum of them and assign as a cost or should we change the structure of the errr itself and instead of cost we will just have appBudgetConsumed and logicSigBudgetConsumed as fields respectively? Cause its no longer dealing with dryrun related outputs and only relying on simulate, might make sense to adjust the error fields to items specific to simulate?

robdmoore: I'd use appBudget for this, and maybe add logic app as a separate one if we want it

src/transaction.ts
aorumbayev: We don't. I've sent @negar-abbasi a snipped of how signatures can be ignored she will push it later on, we will also smoke test it with real wallet just to be sure

src/transaction.ts
```diff
@@ -245,25 +246,25 @@ export const sendAtomicTransactionComposer = async function (atcSend: AtomicTran
...
           e.traces.push({
-            trace: txn.appTrace(),
-            cost: txn.cost,
-            logs: txn.logs,
-            messages: txn.appCallMessages,
+            trace: txn.execTrace,
+            appBudget: txn.appBudgetConsumed,
+            logicSigBudget: txn.logicSigBudgetConsumed,
+            logs: txn.txnResult.logs,
+            messages: simulate.txnGroups[0].failureMessage,
```
neilcampbell, line 263: Previously this was a `string[]`, now it's a `string`.

aorumbayev (approved):

neilcampbell (approved):

PR #134, opened 2023-09-11
feat: transfer assets between accounts

inaie-makerx:
Transfer assets between two addresses

### DIFF
--- a/docs/capabilities/transfer.md
+++ b/docs/capabilities/transfer.md
@@ -38,7 +38,7 @@ The key function to facilitate asset transfers is `transferAsset(transfer, algod
 - `from: SendTransactionFrom` - The account that will send the asset
 - `to: string` - The account / account address that will receive the asset
 - `assetID: number` - The asset id that will be transfered
-- `amount: AlgoAmount` - The amount to send as the smallest divisible unit value
+- `amount: number` - The amount to send as the smallest divisible unit value
 - `transactionParams?: SuggestedParams` - The optional [transaction parameters](./transaction.md#transaction-params)
 - `clawbackFrom: SendTransactionFrom | string` - An optional address of a target account from which to perform a clawback operation. Please note, in such cases senderAccount must be equal to clawback field on ASA metadata.
 - `note?: TransactionNote` - The [transaction note](./transaction.md#transaction-notes)
--- a/docs/capabilities/transfer.md
+++ b/docs/capabilities/transfer.md
@@ -30,15 +30,9 @@ The ability to automatically fund an account to have a minimum amount of disposa
 
 The function calls Algod to find the current balance and minimum balance requirement, gets the difference between those two numbers and checks to see if it's more than the `minSpendingBalance` and if so then it will send the difference, or the `minFundingIncrement` if that is specified.
 
-## Dispenser
-
-If you want to programmtically send funds then you will often need a "dispenser" account that has a store of ALGOs that can be sent and a private key available for that dispenser account.
-
-There is a standard AlgoKit Utils function to get access to a [dispenser account](./account.md#accounts): [`algokit.getDispenserAccount(algod, kmd?)`](../code/modules/index.md#getdispenseraccount). When running against [LocalNet](https://github.com/algorandfoundation/algokit-cli/blob/main/docs/features/localnet.md), the dispenser account can be automatically determined using the [Kmd API](https://developer.algorand.org/docs/rest-apis/kmd). When running against other networks like TestNet or MainNet the mnemonic (and optionally sender address if it's been rekeyed) of the dispenser account can be provided via environment variables (`process.env.DISPENSER_MNEMONIC` and optionally `process.env.DISPENSER_SENDER` if rekeyed).
-
 ## `transferAssets`
 
-The key function to facilitate asset transfers is `transferAsset(transfer, algod)`, which returns a [`SendTransactionResult`](./transaction.md#sendtransactionresult) and takes a [`TransferAssetParams`](../code/interfaces/types_indexer.AssetTransferTransactionResult.md):
+The key function to facilitate asset transfers is `transferAsset(transfer, algod)`, which returns a [`SendTransactionResult`](./transaction.md#sendtransactionresult) and takes a [`TransferAssetParams`](../code/interfaces/types_transfer.TransferAssetParams.md):
 
 - All properties in [`SendTransactionParams`](./transaction.md#sendtransactionparams)
 - `from: SendTransactionFrom` - The account that will send the asset
@@ -48,3 +42,9 @@ The key function to facilitate asset transfers is `transferAsset(transfer, algod
 - `transactionParams?: SuggestedParams` - The optional [transaction parameters](./transaction.md#transaction-params)
 - `clawbackFrom: SendTransactionFrom | string` - An optional address of a target account from which to perform a clawback operation. Please note, in such cases senderAccount must be equal to clawback field on ASA metadata.
 - `note?: TransactionNote` - The [transaction note](./transaction.md#transaction-notes)
+
+## Dispenser
+
+If you want to programmtically send funds then you will often need a "dispenser" account that has a store of ALGOs that can be sent and a private key available for that dispenser account.
+
+There is a standard AlgoKit Utils function to get access to a [dispenser account](./account.md#accounts): [`algokit.getDispenserAccount(algod, kmd?)`](../code/modules/index.md#getdispenseraccount). When running against [LocalNet](https://github.com/algorandfoundation/algokit-cli/blob/main/docs/features/localnet.md), the dispenser account can be automatically determined using the [Kmd API](https://developer.algorand.org/docs/rest-apis/kmd). When running against other networks like TestNet or MainNet the mnemonic (and optionally sender address if it's been rekeyed) of the dispenser account can be provided via environment variables (`process.env.DISPENSER_MNEMONIC` and optionally `process.env.DISPENSER_SENDER` if rekeyed).
--- a/docs/capabilities/transfer.md
+++ b/docs/capabilities/transfer.md
@@ -10,7 +10,7 @@ The key function to facilitate Algo transfers is `algokit.transferAlgos(transfer
 
 - All properties in [`SendTransactionParams`](./transaction.md#sendtransactionparams)
 - `from: SendTransactionFrom` - The account that will send the ALGOs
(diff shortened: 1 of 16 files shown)


src/transfer.ts
```diff
@@ -74,3 +74,53 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
 
   return undefined
 }
+
+interface TransferAssetParams extends SendTransactionParams {
```
robdmoore, line 78: This should do into types/transfer so it can be easily imported when people want to extend this method

src/transfer.ts
```diff
@@ -74,3 +74,53 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
...
+  senderAccount: SendTransactionFrom
```
robdmoore, line 79: As a suggestion I think we should model the type for the `transferAlgos` method i.e. `from`, `to`

src/transfer.ts
robdmoore, line 78: This interface and the method should both have full jsdoc comments for the documentation and intellisense

src/transfer.ts
```diff
@@ -74,3 +74,53 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
...
+  receiverAccount: Account
+  assetId: number
```
robdmoore, line 81: Worth double checking the algosdk parameter, but I suspect this should be `number | bigint` since asset ID is a 64 bit int (admittedly it will take a long time for it to get big enough to not fit into a `number`

src/transfer.ts
```diff
@@ -74,3 +74,53 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
 
   return undefined
 }
+
+interface TransferAssetParams extends SendTransactionParams {
+  senderAccount: SendTransactionFrom
+  receiverAccount: Account
```
robdmoore, line 80: This should be `SendTransactionFrom | string` since we just need the address. See `transferAlgos` to see how it's implemented :)

src/transfer.ts
```diff
@@ -74,3 +74,53 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
...
+  assetId: number
+  amount: number | bigint
+  transactionParams?: SuggestedParams
+  clawbackFrom?: SendTransactionFrom
```
robdmoore, line 84: This should be `SendTransactionFrom | string` since we just need the address so you should be able to pass in a string address without needing the underlying account.

src/transfer.ts
robdmoore: I stand corrected, it's just `number`

src/transfer.ts
```diff
@@ -74,3 +74,53 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
...
+    amount: amount,
+    note: encodeTransactionNote(note),
+    assetIndex: assetId,
+    suggestedParams: await getTransactionParams(transactionParams, algod),
+    rekeyTo: undefined,
+  })
+
+  if (!sendParams.skipSending) {
+    if (sendParams.maxFee === undefined) {
+      sendParams.maxFee = (0.02).algos()
```
robdmoore, line 106: Given we allow peopel to override maxFee I think it's better to not include this because AlgoKit probably shouldn't be opinionated about how much fees people are happy to spend

src/transfer.ts
```diff
@@ -74,3 +74,53 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
...
+    )
+
+    Config.getLogger(sendParams.suppressLog).info(
+      `Successfully transferred ${amount} of asset ${assetId} from ${getSenderAddress(senderAccount)} to ${receiverAccount}.`,
+    )
+
+    return result
+  }
+
+  return { transaction, confirmation: undefined }
```
aorumbayev, line 125: I think reusing the conditional and return statement similar to how it's implemented in transferAlgos will be a better option here.

src/transfer.ts
```diff
@@ -74,3 +74,53 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
 
   return undefined
 }
+
+interface TransferAssetParams extends SendTransactionParams {
+  senderAccount: SendTransactionFrom
+  receiverAccount: Account
+  assetId: number
+  amount: number | bigint
```
neilcampbell, line 82: I really like how `transferAlgos` types `amount` as `AlgoAmount`. It makes it super clear when your dealing with Algos or MicroAlgos and allows users to seamlessly use either.

Obviously transferAsset is a bit more complex, as we don't know what we are dealing with or the dc value for the asset.
I think we should make sure the docs are clear that we expect the smallest unit value, rather than the nominal unit value.

neilcampbell:
@inaie-makerx docs will also need to be updated via `npm run generate:code-docs`, so the `check_docs` action passes.

src/transfer.ts
```diff
@@ -74,3 +74,33 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
 
   return undefined
 }
+
+export async function transferAsset(
+  { from, to, assetId, amount, transactionParams, clawbackFrom, note, ...sendParams }: TransferAssetParams,
+  algod: Algodv2,
+) {
+  const transaction = algosdk.makeAssetTransferTxnWithSuggestedParamsFromObject({
```
aorumbayev, line 82: unit tests are needed for this method too, refer to tests on transfer algos for inspiration. Should cover main edge cases:

- recepient is not opted in - should throw an error 
- clawback is specified, should perform a clawback from target
- assed id is invalid or sender does not hold or opted in to the asset - should throw an error

aorumbayev (changes requested):

src/transfer.ts
```diff
@@ -74,3 +74,33 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
 
   return undefined
 }
+
+export async function transferAsset(
```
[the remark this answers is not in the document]
aorumbayev: Not sure why this one was resolved?
https://github.com/algorandfoundation/algokit-utils-ts/blob/transfer-assets-/src/transfer.ts
JSDocs are still missing

src/transfer.ts
```diff
@@ -74,3 +74,37 @@ export async function ensureFunded(funding: EnsureFundedParams, algod: Algodv2,
...
   return undefined
 }
+
+/**
+ * Transfer asset between two accounts.
+ * @param transfer The transfer definition
+ * @param algod An algod client
+ * @returns The transaction object and optionally the confirmation if it was sent to the chain (`skipSending` is `false` or unset)
+ */
+export async function transferAsset(transfer: TransferAssetParams, algod: Algodv2) {
```
neilcampbell, line 84: This will also need to be documented in the `./docs/capabilities/transfer.md`

aorumbayev (approved):

docs/capabilities/transfer.md
```diff
@@ -35,3 +35,16 @@ The function calls Algod to find the current balance and minimum balance require
 If you want to programmtically send funds then you will often need a "dispenser" account that has a store of ALGOs that can be sent and a private key available for that dispenser account.
 
 There is a standard AlgoKit Utils function to get access to a [dispenser account](./account.md#accounts): [`algokit.getDispenserAccount(algod, kmd?)`](../code/modules/index.md#getdispenseraccount). When running against [LocalNet](https://github.com/algorandfoundation/algokit-cli/blob/main/docs/features/localnet.md), the dispenser account can be automatically determined using the [Kmd API](https://developer.algorand.org/docs/rest-apis/kmd). When running against other networks like TestNet or MainNet the mnemonic (and optionally sender address if it's been rekeyed) of the dispenser account can be provided via environment variables (`process.env.DISPENSER_MNEMONIC` and optionally `process.env.DISPENSER_SENDER` if rekeyed).
+
+## `transferAssets`
```
neilcampbell, line 39: I'd move this above the `Dispenser` section, alongside the other functions.

docs/capabilities/transfer.md
```diff
@@ -35,3 +35,16 @@ The function calls Algod to find the current balance and minimum balance require
...
+
+The key function to facilitate asset transfers is `transferAsset(transfer, algod)`, which returns a [`SendTransactionResult`](./transaction.md#sendtransactionresult) and takes a [`TransferAssetParams`](../code/interfaces/types_indexer.AssetTransferTransactionResult.md):
```
neilcampbell, line 41: `../code/interfaces/types_indexer.AssetTransferTransactionResult.md` doesn't appear to be the correct docs for that type.

docs/capabilities/transfer.md
```diff
@@ -30,6 +30,19 @@ The ability to automatically fund an account to have a minimum amount of disposa
...
 
+## `transferAssets`
+
+The key function to facilitate asset transfers is `transferAsset(transfer, algod)`, which returns a [`SendTransactionResult`](./transaction.md#sendtransactionresult) and takes a [`TransferAssetParams`](../code/interfaces/types_transfer.TransferAssetParams.md):
+
+- All properties in [`SendTransactionParams`](./transaction.md#sendtransactionparams)
+- `from: SendTransactionFrom` - The account that will send the asset
+- `to: string` - The account / account address that will receive the asset
+- `assetID: number` - The asset id that will be transfered
+- `amount: AlgoAmount` - The amount to send as the smallest divisible unit value
```
neilcampbell, line 41: This parameter isn't correct

neilcampbell, line 41: This should be a number

docs/capabilities/transfer.md
```diff
@@ -30,6 +30,19 @@ The ability to automatically fund an account to have a minimum amount of disposa
 
 The function calls Algod to find the current balance and minimum balance requirement, gets the difference between those two numbers and checks to see if it's more than the `minSpendingBalance` and if so then it will send the difference, or the `minFundingIncrement` if that is specified.
 
+## `transferAssets`
+
+The key function to facilitate asset transfers is `transferAsset(transfer, algod)`, which returns a [`SendTransactionResult`](./transaction.md#sendtransactionresult) and takes a [`TransferAssetParams`](../code/interfaces/types_transfer.TransferAssetParams.md):
+
+- All properties in [`SendTransactionParams`](./transaction.md#sendtransactionparams)
```
neilcampbell, line 37: I think this is the wrong type

aorumbayev (approved):

PR #144, opened 2023-09-25
feat: adding support for interacting with dispenser api on testnet via ensureFunded

aorumbayev:
## Proposed Changes

- Refactoring ensureFunded to invoke dispenser api if optional bool set to true. Maintains relative feature parity with implementation on https://github.com/algorandfoundation/algokit-utils-py/pull/51
- Adding pr template
- Changing recommended test runner plugin (the current one has a lot of issues with issuing debug sessions per each individual test)

## Further notes

- Docs needs to be updated to include URL to API docs once @loedn decides on where to host the API docs


.vscode/extensions.json
```diff
@@ -1,3 +1,3 @@
 {
-  "recommendations": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode", "orta.vscode-jest"]
+  "recommendations": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode", "firsttris.vscode-jest-runner"]
```
robdmoore, line 2: Interesting, what does this extension do differently?

src/types/transaction.ts
```diff
@@ -46,6 +46,12 @@ export interface SendTransactionResult {
   confirmation?: modelsv2.PendingTransactionResponse
 }
 
+/** The result of sending a fund transaction from AlgoKit Dispenser API */
+export interface SendDispenserTransactionResult {
+  /** The transaction ID */
+  transaction: string
```
robdmoore, line 52: This should probably be `transactionId` to be clearer

src/types/transfer.ts
```diff
@@ -30,6 +36,8 @@ export interface EnsureFundedParams extends SendTransactionParams {
   transactionParams?: SuggestedParams
   /** The (optional) transaction note, default: "Funding account to meet minimum requirement" */
   note?: TransactionNote
+  /** Whether to use the dispenser API to fund the account. Defaults to false. */
+  useDispenserApi?: boolean
```
robdmoore, line 40: I think `fundingSource` allowing `'testnet-dispenser-api'` as a union type option is a cleaner way of doing it.

src/util.ts
```diff
@@ -20,3 +20,23 @@ export const toNumber = (value: number | bigint) => {
...
 export class UnsafeConversionError extends Error {}
+
+/**
+ * Calculates the amount of funds to add to a wallet to bring it up to the minimum spending balance.
+ * @param minSpendingBalance The minimum spending balance for the wallet
+ * @param currentSpendingBalance The current spending balance for the wallet
+ * @param minFundingIncrement The minimum amount of funds that can be added to the wallet
+ * @returns The amount of funds to add to the wallet or null if the wallet is already above the minimum spending balance
+ */
+export const calculateFundAmount = (
```
robdmoore, line 31: Is this exported into algokit itself? If not then all good, if it is we probably need a better name for the function that's more descriptive.

.vscode/extensions.json
aorumbayev: @robdmoore i did discard that change in the end, facebook's one still has better support for test explorer, the main pain was default config for auto running jest on each save which i got rid of - it was making ide slow and sluggish and i don't see much value in that anyway since CI will prevent non passing tests anyway and i added a custom npm command to simulate `pre-commit`

src/types/transfer.ts
aorumbayev: @robdmoore closing as this is now taking a client instance

src/types/transaction.ts
aorumbayev: closing, this is completely refactored now

src/util.ts
aorumbayev: @robdmoore thats all thats exported, don't see utils here so should be good:

```ts
import { UpdatableConfig } from './types/config'

export * from './account'
export * from './amount'
export * from './app'
export * from './app-client'
export * from './app-deploy'
export * from './indexer-lookup'
export * from './localnet'
export * from './transaction'
export * from './transfer'

/** The AlgoKit config. To update it use the configure method. */
export const Config = new UpdatableConfig()
```

aorumbayev:
@robdmoore ready for review

docs/capabilities/dispenser-client.md
```diff
@@ -0,0 +1,57 @@
...
+1. Pass the token directly to the client constructor as `authToken`.
+2. Set the token as an environment variable `ALGOKIT_DISPENSER_ACCESS_TOKEN` (see [docs](https://github.com/algorandfoundation/algokit/blob/main/docs/testnet_api.md#error-handling) on how to obtain the token).
+
+If both methods are used, the constructor argument takes precedence.
+
+```ts
+import * as algokit from '@algorandfoundation/algokit-utils'
+
+// Using constructor argument
+const client = new algokit.getDispenserApiTestnetClient({ authToken: 'your_auth_token' })
```
robdmoore, line 18: Is the `new` correct here? I would expect that to not need a `new`

docs/capabilities/dispenser-client.md
```diff
@@ -0,0 +1,57 @@
...
+
+// Using environment variable
+process.env['ALGOKIT_DISPENSER_ACCESS_TOKEN'] = 'your_auth_token'
+const client = new DispenserApiTestnetClient()
```
robdmoore, line 22: Should this also use the algokit method rather than the client class directly? And if so should that method be explicit it's using environment in the method name like some of the others?

docs/capabilities/dispenser-client.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+> Keep in mind, to perform a refund you need to perform a payment transaction yourself first by sending funds back to TestNet Dispenser, then you can invoke this refund endpoint and pass the txn_id of your refund txn. You can obtain dispenser address by inspecting the sender field of any issued fund transaction initiated via [fund](#funding-an-account).
+
+## Getting Current Limit
+
+To get the current limit for an account with Algos from the dispenser API, use the `limit` method. This method requires the account address.
+
+```ts
+const response = await client.limit()
```
robdmoore, line 50: This should probably be `getLimit` since we aren't performing a (verb) action

robdmoore (approved):

docs/capabilities/dispenser-client.md
aorumbayev: was just a typo, the constructor within the client by default will throw and error if there is no env and no token supplied so not sure we need a separate method just for env loading since user can just init without passing anything (while making sure env is available, and thats also documented in detail in docstrings)

docs/capabilities/dispenser-client.md
aorumbayev: https://github.com/algorandfoundation/algokit-utils-py/pull/52 this was already limit on py version so i assumed we are ok with the title, adding a pr there too so its get_limit in py and getLimit here

docs/capabilities/dispenser-client.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Registering a Refund
+
+To register a refund for a transaction with the dispenser API, use the `refund` method. This method requires the transaction ID of the refund transaction.
+
+```ts
+await client.refund('transaction_id')
+```
+
+> Keep in mind, to perform a refund you need to perform a payment transaction yourself first by sending funds back to TestNet Dispenser, then you can invoke this refund endpoint and pass the txn_id of your refund txn. You can obtain dispenser address by inspecting the sender field of any issued fund transaction initiated via [fund](#funding-an-account).
```
robdmoore, line 43: In time it would be neat to have the client be able to create and/or issue the transaction, but I think what you have is fine as a starting point

src/types/dispenser-client.ts
```diff
@@ -0,0 +1,179 @@
...
+      method: method,
+      headers: headers,
+      signal: AbortSignal.timeout(this.requestTimeout * 1000),
+    }
+
+    if (body) {
+      requestArgs.body = JSON.stringify(body)
+    }
+
+    const response = await fetch(`${baseUrl}/${urlSuffix}`, requestArgs)
```
robdmoore, line 113: Are there any issues with relying on global fetch? We have minimum of node 18 don't we? Is global fetch available in node 18?

robdmoore (approved):

src/types/dispenser-client.ts
aorumbayev: @robdmoore didn't notice that we use croos-fetch in urltokenbasehttpclient class we already have in utils. Will open a follow up chore pr to address this

PR #148, opened 2023-10-11
feat: task opt in and opt out

negar-abbasi:
## Proposed Changes

- opt in to multiple asset
- opt out on multiple assets, grouped into a single transaction 

## TODO
- [x] Add some docs
- [x] Add multiple assets opt in

### DIFF
--- a/src/asset.ts
+++ b/src/asset.ts
@@ -6,14 +6,17 @@ import { TransactionGroupToSend, TransactionToSign } from './types/transaction'
 const MaxTxGroupSize = 16
 
 export async function optIn(client: Algodv2, account: Account, assetId: number) {
+  let accountInfo
   try {
-    const accountInfo = await client.accountInformation(account.addr).do()
-    if (accountInfo.assets.find((a: Record<string, number>) => a['asset-id'] === assetId)) {
-      throw new Error(`Account ${account.addr} have already opt-in to asset ${assetId}`)
-    }
-  } catch {
-    throw new Error(`Account address supplied does not exist`)
+    accountInfo = await client.accountInformation(account.addr).do()
+  } catch (error) {
+    throw new Error(`Account address ${account.addr} does not exist`)
   }
+
+  if (accountInfo.assets.find((a: Record<string, number>) => a['asset-id'] === assetId)) {
+    throw new Error(`Account ${account.addr} has already opted-in to asset ${assetId}`)
+  }
+
   await transferAsset(
     {
       from: account,
--- a/src/asset.ts
+++ b/src/asset.ts
@@ -27,27 +27,38 @@ function* chunks<T>(arr: T[], n: number): Generator<T[], void> {
     yield arr.slice(i, i + n)
   }
 }
-async function checkAssetBalance(account: Account, assetId: number, client: Algodv2) {
-  const accountInfo = await client.accountInformation(account.addr).do()
-  if (!accountInfo.assets || !accountInfo.assets.find((a: Record<string, number>) => a['asset-id'] === assetId)) {
-    throw new Error(`Account ${account.addr} does not have asset ${assetId}`)
-  }
-  const accountAssetInfo = await client.accountAssetInformation(account.addr, assetId).do()
-  if (accountAssetInfo['asset-holding']['amount'] != 0) {
-    throw new Error(`asset ${assetId} is not with zero balance`)
+async function ensureAssetBalance(account: Account, assetIds: number[], client: Algodv2) {
+  const assetPromises = assetIds.map(async (assetId) => {
+    try {
+      const accountAssetInfo = await client.accountAssetInformation(account.addr, assetId).do()
+      if (accountAssetInfo['asset-holding']['amount'] !== 0) {
+        Config.logger.error(`asset ${assetId} is not with zero balance`)
+        return assetId
+      }
+    } catch (e) {
+      Config.logger.error(`Account ${account.addr} does not have asset ${assetId}`)
+      return assetId
+    }
+    return null
+  })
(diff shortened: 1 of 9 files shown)


src/asset.ts
```diff
@@ -0,0 +1,99 @@
...
+import { TransactionGroupToSend, TransactionToSign } from './types/transaction'
+
+const MaxTxGroupSize = 16
+
+export async function optIn(client: Algodv2, account: Account, assetId: number) {
+  const accountInfo = await client.accountInformation(account.addr).do()
+  if (accountInfo.assets.find((a: Record<string, number>) => a['asset-id'] === assetId)) {
+    throw new Error(`Account ${account.addr} have already opt-in to asset ${assetId}`)
+  }
+  await transferAsset(
```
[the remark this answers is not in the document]
negar-abbasi: It is not hard but the user story is for only one asset to be opted in and a number of assets to be opted out.

src/asset.ts
```diff
@@ -0,0 +1,152 @@
...
+import { Config, sendGroupOfTransactions } from '.'
+import { TransactionGroupToSend, TransactionToSign } from './types/transaction'
+
+const MaxTxGroupSize = 16
+
+function* chunks<T>(arr: T[], n: number): Generator<T[], void> {
+  for (let i = 0; i < arr.length; i += n) yield arr.slice(i, i + n)
+}
+
+async function ensureAccountIsValid(client: algosdk.Algodv2, account: algosdk.Account) {
```
aorumbayev, line 11: Same comment as in `py` version, are these meant to be used only by optin and optout or we want to export them? If we aren't exporting them ensure you are only exposing the optIn and optOut to users

src/asset.ts
```diff
@@ -0,0 +1,152 @@
...
+  try {
+    await client.accountInformation(account.addr).do()
+  } catch (error) {
+    throw new Error(`Account address ${account.addr} does not exist`)
+  }
+}
+
+async function ensureAssetFirstOptIn(client: algosdk.Algodv2, account: algosdk.Account, assetIds: number[]) {
```
aorumbayev, line 19: Same comment as in `py` version, are these meant to be used only by optin and optout or we want to export them? If we aren't exporting them ensure you are only exposing the optIn and optOut to users

src/asset.ts
```diff
@@ -0,0 +1,152 @@
...
+        )
+      })
+    } catch (e) {
+      throw new Error(`Received error trying to opt in ${e}`)
+    }
+  }
+  return result
+}
+
+export async function optOut(client: Algodv2, optoutAccount: Account, assetIds: number[]): Promise<Record<number, string>> {
```
aorumbayev, line 108: in py version you named the account field as `account`, lets either rename it to account here or rename it to match the current title here on py version for consistency

aorumbayev, line 108:
```suggestion
/**
 * Opt out of multiple assets in Algorand blockchain.
 *
 * @param {Algodv2} client - An instance of the Algodv2 client used to interact with the Algorand blockchain.
 * @param {Account} optoutAccount - The Algorand account that wants to opt out of the assets.
 * @param {number[]} assetIds - An array of asset IDs representing the assets to opt out of.
 * @returns {Promise<Record<number, string>>} - A record object containing asset IDs as keys and their corresponding transaction IDs as values.
 */
export async function optOut(client: Algodv2, optoutAccount: Account, assetIds: number[]): Promise<Record<number, string>> {
```

src/asset.ts
```diff
@@ -0,0 +1,152 @@
...
+  if (invalidAssets.length > 0) {
+    throw new Error(
+      `Assets ${invalidAssets.join(
+        ', ',
+      )} cannot be opted out. Ensure that they are valid and that the account has previously opted into them.`,
+    )
+  }
+}
+
+export async function optIn(client: Algodv2, optInAccount: Account, assetIds: number[]) {
```
aorumbayev, line 63:
```suggestion
/**
 * Opt in to a list of assets on the Algorand blockchain.
 *
 * @param client - An instance of the Algodv2 class from the `algosdk` library.
 * @param optInAccount - An instance of the Account class from the `algosdk` library representing the account that wants to opt in to the assets.
 * @param assetIds - An array of asset IDs that the account wants to opt in to.
 * @returns A record object where the keys are the asset IDs and the values are the corresponding transaction IDs for successful opt-ins.
 * @throws If there is an error during the opt-in process.
 */
export async function optIn(client: Algodv2, optInAccount: Account, assetIds: number[]) {
```

aorumbayev (changes requested):
@negar-abbasi please also pull latest from main branch as well as its out of sync by a good # of commits

src/asset.ts
```diff
@@ -0,0 +1,169 @@
...
+
+async function ensureAccountIsValid(client: algosdk.Algodv2, account: algosdk.Account) {
+  try {
+    await client.accountInformation(account.addr).do()
+  } catch (error) {
+    throw new Error(`Account address ${account.addr} does not exist`)
+  }
+}
+
+async function ensureAssetFirstOptIn(client: algosdk.Algodv2, account: algosdk.Account, assetIds: number[]) {
```
aorumbayev, line 19: This method and the one below are redundant, both are almost identical except for min balance parameters, you can combine them into a single method called ensureAssetBalance and take a third param of custom enum type call ValidationType (with option OptIn or OptOut), based on this enum you can them figure out whether to check if balance is >= 0 (opted in) or asset should not be inside accountInfo response at all (opted out)

aorumbayev (changes requested):

aorumbayev (approved):

PR #177, opened 2023-11-29
feat: populate app call resources from algod simulate

joe-p:
## Proposed Changes

- adds `packResources` option to sendParams which will automatically populate app call resource arrays using simulate

## TODO

- [x] Implement tests (tests for this function are available [here](https://github.com/joe-p/resource-packer/blob/main/__test__/resource-packer.test.ts), but can't be easily ported since they use the auto-generated client)
- [x] Adjust simulate error handling to output the same error format as regular appClient errors
- [x] Support resource packing in other methods outside of `sendAtomicTransactionComposer` (ie. `sendTransaction`)

### DIFF
--- a/src/transaction.ts
+++ b/src/transaction.ts
@@ -217,6 +217,160 @@ export const sendTransaction = async function (
 }
 
 /**
+ * Get all of the unamed resources used by the group in the given ATC
+ *
+ * @param algod The algod client to use for the simulation
+ * @param atc The ATC containing the txn group
+ * @returns The unnamed resources accessed by the group and by each transaction in the group
+ */
+export async function getUnnamedResourcesAccessed(algod: algosdk.Algodv2, atc: algosdk.AtomicTransactionComposer) {
+  const simReq = new algosdk.modelsv2.SimulateRequest({
+    txnGroups: [],
+    allowUnnamedResources: true,
+  })
+
+  const result = await atc.simulate(algod, simReq)
+
+  const groupResponse = result.simulateResponse.txnGroups[0]
+
+  if (groupResponse.failureMessage) {
+    throw Error(`Error during resource packing simulation in transaction ${groupResponse.failedAt}: ${groupResponse.failureMessage}`)
+  }
+
+  return {
+    group: groupResponse.unnamedResourcesAccessed,
+    txns: groupResponse.txnResults.map(
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (t: any) => t.unnamedResourcesAccessed,
+    ) as algosdk.modelsv2.SimulateUnnamedResourcesAccessed[],
+  }
+}
+
+/**
+ * Take an existing ATC and return a new ATC with packed resources
+ *
+ * @param algod The algod client to use for the simulation
+ * @param atc The ATC containing the txn group
+ * @returns A new ATC with the resources packed into the transactions
+ */
+export async function packResources(algod: algosdk.Algodv2, atc: algosdk.AtomicTransactionComposer) {
+  const unnamedResourcesAccessed = await getUnnamedResourcesAccessed(algod, atc)
+  const group = atc.buildGroup()
+
+  unnamedResourcesAccessed.txns.forEach((r, i) => {
+    if (r === undefined) return
+
+    if (r.boxes || r.extraBoxRefs) throw Error('Unexpected boxes at the transaction level')
+    if (r.appLocals) throw Error('Unexpected app local at the transaction level')
+    if (r.assetHoldings) throw Error('Unexpected asset holding at the transaction level')
+
+    // Do accounts first because the account limit is 4
+    r.accounts?.forEach((a) => {
+      group[i].txn.appAccounts?.push(algosdk.decodeAddress(a))
+    })
+
+    r.apps?.forEach((a) => {
+      group[i].txn.appForeignApps?.push(Number(a))
+    })
+
+    r.assets?.forEach((a) => {
+      group[i].txn.appForeignAssets?.push(Number(a))
+    })
+
+    const accounts = group[i].txn.appAccounts?.length || 0
+    if (accounts > 4) throw Error(`Account reference limit of 4 exceeded in transaction ${i}`)
+
+    const assets = group[i].txn.appForeignAssets?.length || 0
+    const apps = group[i].txn.appForeignApps?.length || 0
+    const boxes = group[i].txn.boxes?.length || 0
+
+    if (accounts + assets + apps + boxes > 8) throw Error(`Resource reference limit of 8 exceeded in transaction ${i}`)
+  })
+
+  const findTxnBelowRefLimit = (
+    txns: algosdk.TransactionWithSigner[],
+    type: 'account' | 'assetHolding' | 'appLocal' | 'other' = 'other',
+  ) => {
+    const txnIndex = txns.findIndex((t) => {
+      const accounts = t.txn.appAccounts?.length || 0
+      if (type === 'account') return accounts < 4
+
+      const assets = t.txn.appForeignAssets?.length || 0
+      const apps = t.txn.appForeignApps?.length || 0
+      const boxes = t.txn.boxes?.length || 0
+
+      if (type === 'assetHolding' || type === 'appLocal') {
+        return accounts + assets + apps + boxes < 7 && accounts < 4
+      }
+
+      return accounts + assets + apps + boxes < 8
+    })
+
+    if (txnIndex === -1) {
+      throw Error('No more transactions below reference limit. Add another app call to the group.')
+    }
+
+    return txnIndex
+  }
+
+  const g = unnamedResourcesAccessed.group
+
+  if (g) {
+    // Do cross-reference resources first because they are the most restrictive in terms
+    // of which transactions can be used
+    g.appLocals?.forEach((a) => {
+      const txnIndex = findTxnBelowRefLimit(group, 'appLocal')
+      group[txnIndex].txn.appForeignApps?.push(Number(a.app))
+      group[txnIndex].txn.appAccounts?.push(algosdk.decodeAddress(a.account))
+    })
+
+    g.assetHoldings?.forEach((a) => {
+      const txnIndex = findTxnBelowRefLimit(group, 'assetHolding')
+      group[txnIndex].txn.appForeignAssets?.push(Number(a.asset))
+      group[txnIndex].txn.appAccounts?.push(algosdk.decodeAddress(a.account))
+    })
+
+    // Do accounts next because the account limit is 4
+    g.accounts?.forEach((a) => {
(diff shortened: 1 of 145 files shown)


src/transaction.ts
```diff
@@ -216,6 +216,160 @@ export const sendTransaction = async function (
...
+    })
+
+    const accounts = group[i].txn.appAccounts?.length || 0
+    if (accounts > 4) throw Error(`Account reference limit of 4 exceeded in transaction ${i}`)
+
+    const assets = group[i].txn.appForeignAssets?.length || 0
+    const apps = group[i].txn.appForeignApps?.length || 0
+    const boxes = group[i].txn.boxes?.length || 0
+
+    if (accounts + assets + apps + boxes > 8) throw Error(`Resource reference limit of 8 exceeded in transaction ${i}`)
```
aorumbayev, line 287: Can we have a const variable for ref limit or is there an const in algosdk that can be used instead of hardcoded 8?

src/transaction.ts
```diff
@@ -216,6 +216,160 @@ export const sendTransaction = async function (
...
+    r.apps?.forEach((a) => {
+      group[i].txn.appForeignApps?.push(Number(a))
+    })
+
+    r.assets?.forEach((a) => {
+      group[i].txn.appForeignAssets?.push(Number(a))
+    })
+
+    const accounts = group[i].txn.appAccounts?.length || 0
+    if (accounts > 4) throw Error(`Account reference limit of 4 exceeded in transaction ${i}`)
```
aorumbayev, line 281: Can we have a constant variable instead of hardcoded number?

src/transaction.ts
```diff
@@ -216,6 +216,160 @@ export const sendTransaction = async function (
...
+      for (let i = 0; i < g.extraBoxRefs; i += 1) {
+        const txnIndex = findTxnBelowRefLimit(group)
+        group[txnIndex].txn.boxes?.push({ appIndex: 0, name: new Uint8Array(0) })
+      }
+    }
+  }
+
+  const newAtc = new algosdk.AtomicTransactionComposer()
+
+  group.forEach((t) => {
```
aorumbayev, line 364: wouldn't using clone be a bit simpler in this case? Not sure if it auto cleans the group field for you though

aorumbayev: [AtomicTransactionComposer](https://algorand.github.io/js-algorand-sdk/classes/AtomicTransactionComposer.html#clone.clone-1)

src/transaction.ts
```diff
@@ -216,6 +216,160 @@ export const sendTransaction = async function (
...
+}
+
+/**
+ * Take an existing ATC and return a new ATC with packed resources
+ *
+ * @param algod The algod client to use for the simulation
+ * @param atc The ATC containing the txn group
+ * @returns A new ATC with the resources packed into the transactions
+ */
+export async function packResources(algod: algosdk.Algodv2, atc: algosdk.AtomicTransactionComposer) {
```
aorumbayev, line 256: I assume implementation of `packResources` is gonna be replaced by call to an endpoint via algod? Any chance you can also add some todo notes on chunks that will only live temporarily until replaced with appropriate algod endpoint?

aorumbayev (changes requested):
@joe-p  Minor comments/questions

Regarding pipelines, 
You can use the npm run pre-commit script that runs everything we got in ci, tests and regens the docs for you.

joe-p:
I just pushed the tests, but the error handling during simulate needs to be updated

src/transaction.ts
joe-p: I don't think algosdk has those constants but it would really be useful. For now I'll just make constants local to algokit but I think a PR upstream to algosdk with all the consensus parameter tables would be useful.

src/transaction.ts
joe-p: Didn't even know this was a thing... I'll leverage it in the next commit

joe-p: Actually I realized I can't use this because I am mutating the transactions given by `atc.buildGroup()`

src/transaction.ts
joe-p: Added a comment in [1d58463](https://github.com/algorandfoundation/algokit-utils-ts/pull/177/commits/1d5846351db3267f1574d6ff65e1f71fb7316776). Pretty much the entire function will be replaced once its in algod.

src/transaction.ts
```diff
@@ -202,18 +204,193 @@ export const sendTransaction = async function (
...
+ * @privateRemarks
+ *
+ * This entire function will eventually be implemented in simulate upstream in algod. The simulate endpoint will return
+ * an array of refference arrays for each transaction, so this eventually will eventually just call simulate and set the
+ * reference arraysm in the transactions to the reference arrays returned by simulate.
+ *
+ * See https://github.com/algorand/go-algorand/pull/5684
+ *
+ */
+export async function packResources(algod: algosdk.Algodv2, atc: algosdk.AtomicTransactionComposer) {
```
robdmoore, line 277: I wonder if `packResources` is descriptive enough, since it's exported, it'll be available in algokit.packResources.

I think a better name would be something like `packAppCallResources`

src/transaction.ts
```diff
@@ -202,18 +204,193 @@ export const sendTransaction = async function (
...
+    group: groupResponse.unnamedResourcesAccessed,
+    txns: groupResponse.txnResults.map(
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (t: any) => t.unnamedResourcesAccessed,
+    ) as algosdk.modelsv2.SimulateUnnamedResourcesAccessed[],
+  }
+}
+
+/**
+ * Take an existing ATC and return a new ATC with packed resources
```
robdmoore, line 262:
```suggestion
 * Take an existing Atomic Transaction Composer and return a new one with the required
 *  app call resources packed into it
```

src/transaction.ts
robdmoore, line 277: For consistency with the rest of algokit utils `algod` should be the second argument, not the first.

src/transaction.ts
```diff
@@ -225,7 +402,22 @@ export const sendTransaction = async function (
...
+  let atc: AtomicTransactionComposer
+
+  const hasAppCalls = () =>
+    givenAtc
+      .buildGroup()
+      .map((t) => t.txn.type)
+      .includes(algosdk.TransactionType.appl)
+
+  // If packResources is true OR if packResources is undefined and there are app calls, then pack resources
+  if (sendParams?.packResources || (sendParams?.packResources === undefined && hasAppCalls())) {
```
robdmoore, line 416: If it wasn't for this (adding the call to `sendAtomicTransactionComposer`, then I'd suggest adding `packResources` to `CoreAppCallArgs` so it's explicitly tied to app call transactions, but then we can't do this here so on balance `sendParams` is probably the right place for it.

src/types/transaction.ts
```diff
@@ -41,6 +41,8 @@ export interface SendTransactionParams {
   maxFee?: AlgoAmount
   /** The maximum number of rounds to wait for confirmation, only applies if `skipWaiting` is `undefined` or `false`, default: wait up to 5 rounds */
   maxRoundsToWaitForConfirmation?: number
+  /** Whether to use simulate to automatically pack app call resources into the txn objects. Defaults to true when there are app calls in the group.  */
+  packResources?: boolean
```
robdmoore, line 45: I think it should be called `packAppCallResources`

I also wonder if `undefined` should default to `true`, it's a bit of a breaking change in behaviour, but I guess it's low risk assuming that if you correctly specify the resources I assume they aren't returned as `unnamed resources`?

src/transaction.ts
```diff
@@ -202,18 +204,193 @@ export const sendTransaction = async function (
...
+}
+
+/**
+ * Get all of the unamed resources used by the group in the given ATC
+ *
+ * @param algod The algod client to use for the simulation
+ * @param atc The ATC containing the txn group
+ * @returns The unnamed resources accessed by the group and by each transaction in the group
+ */
+export async function getUnnamedResourcesAccessed(algod: algosdk.Algodv2, atc: algosdk.AtomicTransactionComposer) {
```
robdmoore, line 238: I don't think we should export this one, let's keep it internal?

We should also potentially rename to make it clear it's about app call resources?

And algod should be the second argument too.

src/types/transaction.ts
joe-p: Yeah my thinking is that it only changes the behavior of app calls that would've failed anyways.

robdmoore (approved):

joe-p:
Before merging I need to align the error behavior with how apps would typically throw an error. 

Also, @pbennett brought up the fact that `packResources` as a name might be a bit cofnusing, because it implies the user needs to supply the things (resources) getting "packed". Thoughts on `populateResources` instead? Or maybe even a more explicit `populateResourceArrays`

robdmoore:
`populateAppCallResources` get's my vote

joe-p:
So right now most of the failures come from the fact that the error message is now different. I'm not really sure how we can reconcile this without making a breaking change (assuming we consider the changing of an error message a breaking change). I'm also seeing failures during `Deploy app - replace (abi)` test and I'm not sure why.

aorumbayev (changes requested):
@joe-p i found the root cause for failing test you mentioned. Regarding other bits I am don't think (what do you think @neilcampbell @robdmoore) a change in error message should be considered a full blown major version bump in this case  - should be sufficient to update inline snapshots to match new errors expected with resourcePacking (with `npm run test -- -u`, for inline snapshots though you gotta update strings manually). 

Lastly to fix the docs stuff just run `npm run pre-commit` and this will fix it. But i guess for now the main thing is to figure out how to copy the methodCalls as well.

src/transaction.ts
```diff
@@ -202,18 +204,195 @@ export const sendTransaction = async function (
...
+
+  const newAtc = new algosdk.AtomicTransactionComposer()
+
+  group.forEach((t) => {
+    // eslint-disable-next-line no-param-reassign
+    t.txn.group = undefined
+    newAtc.addTransaction(t)
+  })
+
+  return newAtc
```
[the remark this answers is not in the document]
joe-p: Ah interesting ok. I wonder if `.clone()` would help here. I'll play around with it.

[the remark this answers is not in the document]
joe-p: Fixed in [67a18de](https://github.com/algorandfoundation/algokit-utils-ts/pull/177/commits/67a18de42a04db9353238948c98166625884a982). Apparently typescript allows private access via bracket notation

aorumbayev:
@joe-p lets move the pr progress to this branch https://github.com/algorandfoundation/algokit-utils-ts/pull/177 since i wasn't able to push to your forked branches. 

@Loedn confirmed that there should be tweaks in access control, feel free to create branches directly on the repo going forward - this will simplify collaborating on cases when multiple parties need to push to particular branch.

PR #187, opened 2023-12-19
fix: add an isomorphic __dirname implementation

neilcampbell:
Fixes a `ReferenceError: __dirname is not defined in ES module scope` when using `utils-ts` in an ESM project.

This is a temporary fix to solve the issue, however the longer term solution is to move to something like `rollup` or `tsup` where we can inject shims.


aorumbayev (approved):
@neilcampbell just in case would be great if you can also smoke test it against generator ts as well to see if there are no affected tests

aorumbayev (approved):

Issue #193, opened 2024-01-08
Building Next.js 14 project fails with "Module not found: Can't resolve 'fs'" error


No-Cash-7970:
### Subject of the issue

Whenever I try to build [my project](https://github.com/No-Cash-7970/txnDuck/tree/2d8f682cf0fbcf746440487eeed2f0ec9f1607f7) with algokit-utils-ts version v5.1.0 or higher, it fails with an error like the following:
```
⨯ ./node_modules/@algorandfoundation/algokit-utils/esm/debugging.js:144:0
Module not found: Can't resolve 'fs'

https://nextjs.org/docs/messages/module-not-found

Import trace for requested module:
./node_modules/@algorandfoundation/algokit-utils/esm/transaction.js
./node_modules/@algorandfoundation/algokit-utils/esm/index.js
./src/app/lib/txn-data/data-utils.ts
./src/app/lib/txn-data/index.ts
./src/app/[lang]/txn/compose/components/ComposeForm.tsx
```

This error is repeated multiple times. At least once for every component where I import and use algokit-utils. Version 5.0.1 works fine. Algokit-utils are only imported into client components where Node modules like `fs` and `path` are not available.

### Your environment

* Next.js 14.0.4
* algosdk 2.7.0
* TypeScript 5.3.3
* Yarn 4.0.2
* Node 20.9.0
* Windows 11

### Steps to reproduce

These steps are specific to my project
1. `git clone https://github.com/No-Cash-7970/txnDuck.git`
2. `yarn install:dev`
3. `yarn up @algorandfoundation/algokit-utils@5.1.3`
4. `yarn build`

### Expected behaviour

Project builds with latest algokit-utils-ts without error.

### Actual behaviour

Project fails to build with algokit-utils-ts v5.1.0 or higher.

Issue #188, opened 2023-12-19
Build dist output using rollup or tsup


neilcampbell:
We've needed to put a few workarounds in place to produce a cjs and mjs compatible package. These workarounds could be simplified by using a bundler like rollup or tsup.

Workarounds
- Using tsc-alias to rewrite the esm imports to use full paths
- Shimming __dirname usage inline

Additionally we'd (potentially) like to produce a [CDN host-able web bundle](https://github.com/algorandfoundation/algokit-utils-ts/issues/119).

PR #198, opened 2024-01-09
chore: adding note on next js compatibility

aorumbayev:
## Proposed Changes

- Documenting a patch for the bug report https://github.com/algorandfoundation/algokit-utils-ts/issues/193
- Readme describes a current workaround for next js specific behaviour in regards to inability to ignore dynamically imported node specific modules in a browser environment. 
- This will no longer be an issue once we integrate with rollup or a similar bundler, which will allow us to exclude/hot-swap debug util helpers with browser compatible stubs https://github.com/algorandfoundation/algokit-utils-ts/issues/188

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This library follows the [Guiding Principles of AlgoKit](https://github.com/algo
 
 ## NextJS compatibility
 
-`algokit-utils-ts` has a set of `node` specific utilities used for simplifying aggregation of artifacts for [AlgoKit VSCode Debugger Extension](https://github.com/algorandfoundation/algokit-avm-vscode-debugger). Which causes next based projects to fail on `fs` module not found. To fix this issue, you can add the following to your `next.config.js` file:
+`algokit-utils-ts` has a set of `node` specific utilities used for simplifying aggregation of artifacts for [AlgoKit VSCode Debugger Extension](https://github.com/algorandfoundation/algokit-avm-vscode-debugger). Which causes Next.js based projects to fail on `fs` module not found. To fix this issue, you can add the following to your `next.config.js` file:
 
 ```js
   webpack: (config, { isServer }) => {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,30 @@ npm install @algorandfoundation/algokit-utils
 
 This library follows the [Guiding Principles of AlgoKit](https://github.com/algorandfoundation/algokit-cli/blob/main/docs/algokit.md#guiding-principles).
 
+## NextJS compatibility
+
+`algokit-utils-ts` has a set of `node` specific utilities used for simplifying aggregation of artifacts for [AlgoKit VSCode Debugger Extension](https://github.com/algorandfoundation/algokit-avm-vscode-debugger). Which causes next based projects to fail on `fs` module not found. To fix this issue, you can add the following to your `next.config.js` file:
```
neilcampbell, line 25:
```suggestion
`algokit-utils-ts` has a set of `node` specific utilities used for simplifying aggregation of artifacts for [AlgoKit VSCode Debugger Extension](https://github.com/algorandfoundation/algokit-avm-vscode-debugger). Which causes Next.js based projects to fail on `fs` module not found. To fix this issue, you can add the following to your `next.config.js` file:
```

README.md
```diff
@@ -20,6 +20,30 @@ npm install @algorandfoundation/algokit-utils
...
+  },
+```
+
+The root cause is due to the fact that, unlike many frameworks, Next.js allows you to import server-only (Node.js APIs that don't work in a browser) code into your page files. When Next.js builds your project, it removes server only code from your client-side bundle by checking which code exists inside one any of the following built-in methods (code splitting):
+
+- getServerSideProps
+- getStaticProps
+- getStaticPaths
+
+The Module not found: can't resolve 'xyz' error happens when you try to use server only code outside of these methods. Despite `algokit-utils` lazy loading the node specific code dynamically, next js does not seem to correctly identify whether dynamic import is specific to server or client side. Hence the above fix disables the fallback for `fs` module so it ignores polyfilling it on client side.
```
neilcampbell, line 43:
```suggestion
The Module not found: can't resolve 'xyz' error happens when you try to use server only code outside of these methods. Despite `algokit-utils` lazy loading the node specific code dynamically, Next.js does not seem to correctly identify whether a dynamic import is specific to server or client side. Hence the above fix disables the fallback for `fs` module so it ignores polyfilling it on client side.
```

neilcampbell (approved):

neilcampbell:
Based on this, it probably makes sense to bring the rollup/tsup minor enhancement work up in priority.

PR #199, opened 2024-01-12
feat: resource packing (from @joe-p)

aorumbayev:
Moving description from https://github.com/algorandfoundation/algokit-utils-ts/pull/177 from @joe-p 's PR

---

## Proposed Changes

- adds `packResources` option to sendParams which will automatically populate app call resource arrays using simulate

## TODO

- [x] Implement tests (tests for this function are available [here](https://github.com/joe-p/resource-packer/blob/main/__test__/resource-packer.test.ts), but can't be easily ported since they use the auto-generated client)
- [x] Adjust simulate error handling to output the same error format as regular appClient errors
- [x] Support resource packing in other methods outside of `sendAtomicTransactionComposer` (ie. `sendTransaction`)


aorumbayev:
@joe-p i moved your branch, merged latest docs and added few patches for tests. One thing was primarily wrapping the resource packing under the try catch block present on that method to ensure thrown error also returns trace field. Other than that its just updating the inline snapshots. Feel free to commit to this branch if you got anything else otherwise you can approve and will merge so you can play around with the pre release version (in case any follow tweaks would be needed we can address separately). 

Regarding propagating this to python, i assume ts was a more pressing matter so is this a safe assumption to await for algod endpoint availability before we migrate the simliar logic to python utils? (cc @robdmoore )

joe-p (approved):

joe-p:
> Regarding propagating this to python, i assume ts was a more pressing matter so is this a safe assumption to await for algod endpoint availability before we migrate the simliar logic to python utils? (cc @robdmoore )

I don't know when I'm going to start on the `algod` endpoint. The main reasons I chose the TS lib is because its more popular and having this available changes some of the typing in TEALScript to encourage devs to avoid using reference types.

joe-p:
I will say though the plan for algod is basically to implement this exact same algorithm in the simulate endpoint. I imagine migration to python won't be too difficult and once we have the endpoint changes then migrating to that will be trivial

aorumbayev:
@joe-p sounds good lets sync up next week if you hit any issues with pre release - will add an item on python feature parity with this on our backlog- merging pr now given that Rob approved on previous instance already

PR #209, opened 2024-01-21
fix: transactions signed twice with resource population

joe-p:
## Proposed Changes

Modifies the ATC used for simulate when populating resources to use an empty signer. This avoids users being prompted twice for a signature when using an actual wallet.


joe-p:
Live view of me vs typedoc

image

PR #213, opened 2024-01-21
feat: disable resource population by default

joe-p:
## Proposed Changes

Disables resource population by default due to https://github.com/algorand/go-algorand/issues/5914

This is technically a breaking change since it changes the default behavior of atc and appclient calls


joe-p:
Actually might not be an algod bug... need some more time to debug

joe-p:
Ok yep false alarm, not an algod bug. Closing this PR and will open a new one when I figure out the cause

joe-p:
Nope it is indeed an algod bug see https://github.com/algorand/go-algorand/issues/5914

PR #215, opened 2024-01-22
[RFC] Added potential example of how AlgoKit Utils could evolve

robdmoore:
…to have a simpler intellisense experience that exposes the core stuff without as much configurability and flexibility

### DIFF
--- a/src/algokit.ts
+++ b/src/algokit.ts
@@ -1,0 +1,294 @@
+import algosdk from 'algosdk'
+import { assetOptIn, assetOptOut } from './asset'
+import {
+  getAlgoClient,
+  getAlgoIndexerClient,
+  getAlgoKmdClient,
+  getAlgoNodeConfig,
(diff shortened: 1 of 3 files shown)


src/algokit.ts
```diff
@@ -0,0 +1,294 @@
...
+export type AccountAddress = string
+
+export type TransactionSent = {
+  /** The transaction */
+  transaction: Transaction
+  /** The response if the transaction was sent and waited for */
+  confirmation: modelsv2.PendingTransactionResponse
+}
+
+export class AlgoKit {
```
robdmoore, line 40: This isn't what the final code would look like, it's a quick and dirty single file prototype

joe-p:
Love it. This is exactly what I had in mind when I was talking about instantiated algokit client. Basically bringing the appClient experience to all chain interactions.

PR #224, opened 2024-02-09
fix: update logic error parsing to handle the tweaked format from algod

### LINKED ISSUE #223: Support modified LogicError message in algod 3.22
### Subject of the issue

Version 3.22 of algod modified the error message when a LogicError occurs in a transaction, this change means the regex used to identify LogicErrors and extract the program count no longer works

### Steps to reproduce

1. Run `algokit localnet reset --update` This will get the latest algod image
2. Run tests

### Expected behaviour

Tests pass

### Actual behaviour

Tests fail

neilcampbell:
Fixes #223 

Unfortunately we can't use the new structured data payload, as it doesn't contain transactionId or the error message, so we need to parse the string anyway.

### DIFF
--- a/docs/code/modules/index.md
+++ b/docs/code/modules/index.md
@@ -237,7 +237,7 @@ algokit.bulkOptOut({ account: account, assetIds: [12345, 67890] }, algod)
 
 #### Defined in
 
-[src/asset.ts:236](https://github.com/algorandfoundation/algokit-utils-ts/blob/main/src/asset.ts#L236)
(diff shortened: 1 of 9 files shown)


aorumbayev (approved):
@neilcampbell looks good but i see some prettier linting warning in the pipeline, worth fixing those, i think pre-commit script in package.json should handle formatting those unless prettier got out of sync with eslint configs

aorumbayev:
also worth checking tests on generator ts against this version

Issue #175, opened 2023-11-28
Further investigate imports from algokit utils on cjs failing to resolve module when running eslint 


aorumbayev:
Reported by @joe-p , imports of AppClient from utils types path fails during eslint module resolution on https://github.com/algorandfoundation/TEALScript/pull/244/commits/1fe51cdb2e7059afb606c5215584a395c6b2c950. 

Current fix suggested is to apply -> https://www.npmjs.com/package/eslint-config-airbnb-typescript given that airbnb plugin seems to have certain issues when it comes to module resolution and typescript.

neilcampbell:
We are currently converting the project to bundle using Rollup (via #188), and we've tested a configuration which allows us to fix this issue.

neilcampbell:
@joe-p We have updated the bundle to fix this issue you were having. Please update to `5.4.0`

PR #225, opened 2024-02-12
Bundle with rollup

### LINKED ISSUE #188: Build dist output using rollup or tsup

PatrickDinh:
## Proposed Changes

- Build outputs for CommonJS and ESModule with [rollup](https://rollupjs.org/)
- Break big `.ts` files into smaller ones to avoid circular dependency.

Resolves #188, #175


src/localnet/get-or-create-kmd-wallet-account.ts
```diff
@@ -0,0 +1,68 @@
...
+    `LocalNet account '${walletAccount.name}' doesn't yet exist; created account ${account.addr} with keys stored in KMD and funding with ${
+      walletAccount.fundWith?.algos ?? 1000
+    } ALGOs`,
+  )
+
+  // Fund the account from the dispenser
+  await transferAlgos(
+    {
+      amount: walletAccount.fundWith ?? AlgoAmount.Algos(1000),
+      from: await getLocalNetDispenserAccount(algod, kmd),
```
PatrickDinh, line 61: This was changed from 
```
from: await getDispenserAccount(algod, kmd),
```
to avoid circular dependency. I think that we can call `getLocalNetDispenserAccount` directly here because this method is intended for local dev.

src/index.ts
```diff
@@ -1,18 +1,14 @@
...
 export * from './debugging'
 export * from './dispenser-client'
 export * from './indexer-lookup'
 export * from './localnet'
 export * from './network-client'
 export * from './transaction'
 export * from './transfer'
-
-/** The AlgoKit config. To update it use the configure method. */
-export const Config = new UpdatableConfig()
```
PatrickDinh, line 18: This line was moved to `config.ts` to avoid circular dependency.

rollup.config.ts
```diff
@@ -0,0 +1,76 @@
+import nodeResolve from '@rollup/plugin-node-resolve'
```
robdmoore, line 1: Checking: the `.spec.ts` files don't get picked up right?

src/account/get-dispenser-account.ts
```diff
@@ -0,0 +1,22 @@
+import algosdk from 'algosdk'
+import { getLocalNetDispenserAccount } from '../localnet/get-localnet-dispenser-account'
+import { isLocalNet } from '../localnet/is-localnet'
+import { DISPENSER_ACCOUNT } from '../types/account'
+import { getAccount } from './get-account'
+
+import Algodv2 = algosdk.Algodv2
```
robdmoore, line 7: Now we're using rollup, do we still need this weird format for importing from algosdk?

robdmoore (approved):

robdmoore:
Are there any breaking changes in imported file paths likely to affect consumers?

rollup.config.ts
PatrickDinh:
```
"exclude": ["src/**/*.spec.ts", "tests/**/*.*", "src/testing/asset.ts"]
```
We exclude them in `tsconfig.build.json`

PatrickDinh:
> Are there any breaking changes in imported file paths likely to affect consumers?
I tested this change with `generator-ts` and didn't have any issues so far. I will test this change with few more projects.

When I refactored the `.ts` files, I made sure that I move them into a folder and re-export everything with `index.ts`. I hope that it won't cause breaking changes.

src/account/get-dispenser-account.ts
neilcampbell: Good question. Let me check.

neilcampbell: I have confirmed that we can't remove this. The reason we had to make this change was due to algosdk exporting a malformed ESM output, so we need to wait until v3 is released, where this issues is fixed.

See https://github.com/algorandfoundation/algokit-utils-ts/blob/main/.eslintrc#L11 for an explanation.

neilcampbell (approved):

PR #228, opened 2024-02-16
feat: add populateAppCallResources to Config

joe-p:
## Proposed Changes

- adds `populateAppCallResources` to `Config` so it can be easily enabled/disabled by default


src/transaction/transaction.ts
```diff
@@ -207,10 +207,11 @@ export const sendTransaction = async function (
 
   let txnToSend = transaction
 
+  const resroucePopEnabled = sendParams?.populateAppCallResources ?? Config.populateAppCallResources
```
robdmoore, line 210: Misspelt variable. Suggest `populateResources`?

robdmoore (approved):

src/transaction/transaction.ts
joe-p: Done in c07065e48b842f59aec1102add565809144dada4

PR #234, opened 2024-03-07
feat: client and composer

joe-p:
This PR adds two new classes, `AlgokitComposer` and `AlgokitClient`. The core idea is to provide a smooth experience for all chain interactions as discussed in #215. In most cases users are not expected to use the composer directly, but rather through the client. The composer, however, is still exported in case someone wants to use it without the client.

# Problems Addressed

## Mixing of algosdk and algokit-utils

Sending transactions currently requires mixing the usage of algosdk and algokit-utils. 

### Solution

`AlgokitClient` eliminates need for algosdk usage for chain interactions. `AlgokitComposer` is also able to take an `algosdk.AtomicTransactionCompsoer` and add it to its group. 

## Confusing interfaces

Currently, a lot of the interface on algokit-utils can be rather confusing. There are many union types and many ways to do the same thing. For example, in some places when passing a tranasction you use `{txn, signer}`, `{tranasction, signer}`, or `{transaction, from}`. `signer` and `from` can be many different types, inlcuding but not limited to the types used by `signer` or `from` fields in the SDK.

### Solution

`AlgokitComposer` has straight-forward interfaces for all actions. There are no union types and the functions are not clever. There is one way to do a specific action and it is always done that way.

## A lot of async functions

Because of `suggestedParams`, async calls may currently required when forming transactions. This is paticularly annoying when using `.compose()` on an generated app client. Because the chain of function calls require `await` in the middle making it a bit annoying to work with.

### Solution

Nothing async happens until `AlgokitCompose.execute()` or `AlgokitComposer.buildGroup()` 

## Repetitive or old suggestedParams

Suggested params is currently needed for every transaction formation. A common solution is to just get suggested params once, but its possible that the suggested params could be out of date.

### Solution

`AlgokitComposer` caches `suggestedParams` and only does the API call after a certain amount of time has elapsed

## Repetitive signers

For every transaction, there needs to be a way to sign it. Currently there are many different ways to sign transactions with algokit and often you use a `signer` function in multiple places. 

### Solution

`AlgokitClient` allows you to specify a defaultSigner to use for all transactions (useful for wallet interaction). You can also set a unique signer per address (useful for localnet or custodied accounts). Attaching a unique signer to a tranasction is still possible if necessary. 

## Improper fee setting

Many dApps currently do not set fees properly. For example, if an app wants to add extra fees to their current transaction it is common practice to do `(1 + extra_txns) * MIN_FEE`. This is problematic because it does not account for fee scaling.

### Solution (Partial)

`AlgokitComposer` transactions have two fee fields: `flatFee` for setting the flatFee on the transaction and `extraFee` for adding additional microAlgos to the **suggested** fee. This is important because it still takes into account the `feePerByte` recommended by algod for the given transaction. This means you can simple set `extraFee` to `numInners * MIN_FEE` to ensure the fee will always be correct for app calls with inner transactions, even under fee scaling.

### TODO

In the future we need to devise a smart way to handle scenarios where one outer transaction might want to cover the fee for another out transaction. If `feePerByte` is non-zero, then this is hard to calculate. 

## Differing validity window and wait times

It is currently common practice to use the default validity window (1000 rounds) on transactions but only wait a short amount of rounds for confirmation (typically 3-5). This can lead to a confusing UX and potentially double transactions if the user is not aware of the validty windows of their transactions in the event of longer confirmation times.

### Solution

`AlgokitComposer.defaultValidityWindow` is set to a much more reasonable window (for user-facing dApps) of 10 rounds (~30 seconds). By default, `AlgokitComposer.execute()` will wait until the last valid round in the transction group has passed to ensure tranasctions are confirmed when the user is not expecting it.

## ABI method argument composability

Currently, if an application expects another ABI method call as an argument there is no great way to handle that other than manually building the ATC and deconstructing the group. Even after doing that, the return values for the argument calls will not be availible.

### Solution

`AlgokitComposer` method calls support other method calls as arguments.


src/client.ts
```diff
@@ -0,0 +1,84 @@
...
+      return this.cachedSuggestedParams.params
+    }
+
+    const params = await this.algod.getTransactionParams().do()
+    this.cachedSuggestedParams = { params, time: Date.now() }
+
+    return params
+  }
+
+  newGroup() {
```
tristanmenzel, line 41: Should we call this `.compose()` to match the generated typed client?

src/composer.ts
```diff
@@ -0,0 +1,564 @@
+import algosdk from 'algosdk'
+import { sendAtomicTransactionComposer } from './transaction/transaction'
+
+export type CommonTxnParams = {
+  sender: string
+  signer?: algosdk.TransactionSigner
+  rekeyTo?: string
+  note?: Uint8Array
+  lease?: Uint8Array
+  /** The tranasction fee. In most cases you want to use `extraFee` unless setting the fee to 0 to be covered by another transaction */
```
tristanmenzel, line 10:
```suggestion
  /** The transaction fee. In most cases you want to use `extraFee` unless setting the fee to 0 to be covered by another transaction */
```

src/client.ts
joe-p: I intentionally went with `newGroup` because it's self descriptive. Anyone reading the code will deduce that executing this function will start a new group. `compose` on the other hand is not immediately obvious unless you are familiar with the existing terminology used by the generated clients and ATC.

src/composer.ts
```diff
@@ -0,0 +1,564 @@
...
+
+  private buildAtc(
+    atc: algosdk.AtomicTransactionComposer,
+    txnWithSigners: algosdk.TransactionWithSigner[],
+    methodCalls: Map<number, algosdk.ABIMethod>,
+  ) {
+    const currentLength = txnWithSigners.length
+    const group = atc.buildGroup()
+
+    const atcMethodCalls = atc['methodCalls'] as Map<number, algosdk.ABIMethod>
```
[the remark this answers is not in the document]
joe-p: Yeah makes sense to me. PR here https://github.com/algorand/js-algorand-sdk/pull/860

joe-p:
@robdmoore as we discussed I added the `.send` and `.transaction` interface. I also implemented a `maxFee` mechanism. I know it's duplicating whats done at a lower level, but properly integrating all the fee features at a lower level would require a good amount of refactoring so I think it's just best to leave it as is now and we can refactor later. I just want to get user-facing stuff out the door ASAP. 

Only major thing I can think of that we discussed is supporting string on lease and note fields, which shouldn't be too hard. Then testing and docs.

joe-p:
FYI @robdmoore in [b01bd65](https://github.com/algorandfoundation/algokit-utils-ts/pull/234/commits/b01bd658a132fd6e4886e01deb4c67e2141818e1) I changed the algorand client used in the test fixture to be persistent. This is particularly useful when you want to use an account across multiple tests and preserve the signer.

src/types/composer.ts
```diff
@@ -0,0 +1,673 @@
...
+  decimals?: number
+  /** Whether the asset is frozen by default in the creator address */
+  defaultFrozen?: boolean
+  /** The address that can change the manager, reserve, clawback, and freeze addresses */
+  manager?: string
+  /** The address that holds the uncirculated supply */
+  reserve?: string
+  /** The address that can freeze the asset in any account */
+  freeze?: string
+  /** The address that can clawback the asset from any account */
```
tristanmenzel, line 57: Should we include something in these comments to explain that setting them to "" will disable the associated behaviour for the asset?

src/types/composer.ts
```diff
@@ -0,0 +1,673 @@
...
+  /** The default transaction validity window */
+  defaultValidityWindow = 10
+
+  constructor({
+    algod,
+    getSigner,
+    getSuggestedParams,
+    defaultValidityWindow,
+  }: {
+    algod: algosdk.Algodv2
```
tristanmenzel, line 210: Maybe we can use a named type here for the constructor args so we can add comments to the various properties

src/types/account-manager.ts
```diff
@@ -0,0 +1,253 @@
...
+
+/** Creates and keeps track of signing accounts against sending addresses. */
+export class AccountManager {
+  private _clientManager: ClientManager
+  private _accounts: { [address: string]: TransactionSignerAccount } = {}
+  private _defaultSigner?: algosdk.TransactionSigner
+
+  /**
+   * Create a new account creator
+   * @param algod The algod client to use
```
tristanmenzel, line 28: docs don't match the code

src/types/account-manager.ts
```diff
@@ -0,0 +1,253 @@
...
+   * const rekeyedAccount = await account.fromMnemonic("mnemonic secret ...", "SENDERADDRESS...")
+   * ```
+   * @param mnemonicSecret The mnemonic secret representing the private key of an account; **Note: Be careful how the mnemonic is handled**,
+   *  never commit it into source control and ideally load it from the environment (ideally via a secret storage service) rather than the file system.
+   * @param sender The optional sender address to use this signer for (aka a rekeyed account)
+   * @returns The account
+   */
+  public fromMnemonic = (mnemonicSecret: string, sender?: string) => {
+    const account = mnemonicAccount(mnemonicSecret)
+    return this.signerAccount(sender ? rekeyedAccount(account, sender) : account)
```
tristanmenzel, line 143: I don't think `this` will work as `this` won't be bound to the class instance in an arrow function.

tristanmenzel: Hmmm, apparently it does work but only because the assignment is hoisted to the constructor. TIL!

src/types/composer.ts
joe-p: Done in cd4f67278696c74f9aeb2a9852c56f8f774b642a

src/types/composer.ts
joe-p: Done in a94cf05e8db519f01a34c2c5e07efcb66e3010cb

src/types/account-manager.ts
joe-p: Fixed in 05bb21ecb117500bc87be6fe2dd5d831e8ad413f

PR #237, opened 2024-03-12
fix: export all types in types folder

PatrickDinh:
## Proposed Changes

There was an issue with the rollup config in which only two files:
- src/index.ts
- src/testing/index.ts

were included as input for rollup. Rollup only traced references from these two files to generate the output bundle.

In the types folder, there were exported types that weren't referred to by either of the two inputs. These types weren't bundled into the output.

This PR aims to include all files in the types folder into the bundle output.

## How did I test it
- Produced the build output from `main`.
- Added the bundle to an empty git repo and committed.
- Applied the changes from this PR.
- Built and copied the bundle to the above git repo.
- I confirmed that the only changes were:
  - Some internal type names.
  - The missing types were added.
- I also successfully tested this change with `algokit-subscriber`. This is the repo that has the missing types issue.

**Bonus**: Upgrade `semantic-release` to address npm audit issue. 
- The vulnerable package was [ip](https://www.npmjs.com/package/ip) but I couldn't get `overrides` field in `package.json` to work. I went ahead and update `semantic-release`

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -1,7 +1,1 @@
-{
-  "1096519": {
-    "active": true,
-    "notes": "Ignored since the vulnerability is when hosting a server, but this is via semantic release which only uses it as a client",
-    "expiry": 1711814399000
-  }
-}
+{}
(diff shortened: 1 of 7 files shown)


neilcampbell (approved):

PatrickDinh:
> Hey @PatrickDinh, just curious, the readme of that rollup plugin references some rollup specific functionality like output.preserveModulesRoot and output.preserveModules, were those indeed not sufficient in our case?
> 
> CleanShot 2024-03-12 at 09 04 06

We already use `preserveModules = true` so that each module is package separately. The default behaviour is that everything is bundled into 1 file. 

The issue that we are facing here is different ... there were files ignored by rollup because it wasn't referenced by the inputs.

PR #242, opened 2024-03-12
fix: double addr in resource population

joe-p:
## Proposed Changes

- Resource population would erroneously try to put the same address in the account array twice if that address was used for an assetHolding. This would result in a false error regarding not having enough space to populate resources


joe-p:
FYI there seems to be degredation in GitHub PRs right now. The branch is up to date but this PR is not so its missing the docs (which I squashed on the branch as seen here https://github.com/algorandfoundation/algokit-utils-ts/compare/main...fix/resource_pop_double_addr)

robdmoore (approved):

robdmoore:
Weird, I tried closing and reopening and it won't let me reopen

PR #244, opened 2024-03-13
fix: more efficient resource population

joe-p:
## Proposed Changes

- Builds on #243 by checking to see if an account, asset, or app is already available in a transaction so it can be smart about where to place asset holdings/app locals


robdmoore (approved):

SilentRhetoric:
Am I correct to assume that duplicate blank refs for extra box data budget will be preserved?

joe-p:
> Am I correct to assume that duplicate blank refs for extra box data budget will be preserved?

Yes any references that are manually included in the transactions prior to resource population will be preserved. Resource population takes care of extra empty box refs though, so you shouldn't need to manually put them in.

https://github.com/algorandfoundation/algokit-utils-ts/blob/e75765fb693cad746e1399252ae0eedc51fb41c1/src/transaction/transaction.ts#L481-L486

PR #248, opened 2024-03-16
Tweaks to AlgorandClient:

robdmoore:
* Rename AlgoKitClient to AlgorandClient and added static construction methods
* Added fluent with* methods to AlgorandClient to allow for fluent configuration
* Extracted client related stuff to ClientManager and account stuff to AccountManager and added extra functionality to both
* Add transaction and confirmation to the result of sending a single transaction for better ergonomics
* Added ability to specify suggested params and create a copy of suggested params when providing it
* Moved classes to types directory
* Added getAccountInformation to make tests terser (this method should have been added ages ago)
* Incorporated client into testing fixture
* Changed all possible bigint | numbers' to number (where it's impractical for them to hit 2^53)
* Incorporated TransactionSignerAccount with TransactionSigner to allow for terser code
* Rename ID to Id for consistency with rest of algokit utils

There is a txn dead error in the tests to do with valid round calculation that needs to be resolved and I haven't made any changes to composer yet, but there are some I want to suggest, this covers everything else though.


src/testing/fixtures/algorand-fixture.ts
```diff
@@ -92,21 +93,29 @@ export function algorandFixture(fixtureConfig?: AlgorandFixtureConfig, config?:
...
-        transactionLoggerAlgod,
-        kmd,
-      ),
-      generateAccount: (params: GetTestAccountParams) => getTestAccount(params, transactionLoggerAlgod, kmd),
+      testAccount,
+      generateAccount: async (params: GetTestAccountParams) => {
+        const account = await getTestAccount(params, transactionLoggerAlgod, kmd)
+        algorandClient.withAccount(account)
+        return { ...account, signer: algorandClient.account.getSigner(account.addr) }
+      },
```
robdmoore, line 118: This means generated accounts in testing will automatically be registered with the AlgorandClient that is now also available in the fixture.

src/types/account-manager.ts
```diff
@@ -0,0 +1,218 @@
...
+    this._defaultSigner = 'signer' in signer ? signer.signer : signer
+    return this
+  }
+
+  /**
+   * Records the given account against the address of the account for later retrieval and returns a `TransactionSignerAccount`.
+   * @param account The account to use.
+   * @returns A `TransactionSignerAccount` for the given account.
+   */
+  private signerAccount<T extends SendTransactionFrom>(account: T): TransactionSignerAccount & { account: T } {
```
robdmoore, line 38: This is a private method that provides a bridge between the underlying AlgoKit account methods that return various subsets of `SendTransactionFrom` and the consistent `TransactionSignerAccount` type this class exposes. It still provides the underlying type via `account` just in case you want to do stuff with that (e.g. pull out `sk` etc.).

joe-p:
All looks good to me but ditching `bigint` does make me nervous. The SDK is going to all bigint for good reason imo. The moreI think about I think having just `bigint` for all the numeric types makes the most sense. Most basic reasoning being that it's more representative of the actual data type. Also, while it's true that given how things currently work it's unlikely we'll need it, but there could be changes in the future that would end up breaking deployed code. Also there could be forks of Algorand that would want to use this library. 

> There is a txn dead error in the tests to do with valid round calculation that needs to be resolved

Presumably you fixed this? Tests seem to be passing for me.

robdmoore:
bigints are painful to work with, my vote is for using number for things we know will never be > 2^53, but happy to follow consensus from broader group of stakeholders.

The test error I get is `URLTokenBaseHTTPError: Network request error. Received status 400 (Bad Request): TransactionPool.Remember: txn dead: round 418 outside of 397--407`. As far as I can tell a transaction within a group is getting conflicting rounds.

joe-p:
> bigints are painful to work with, my vote is for using number for things we know will never be > 2^53, but happy to follow consensus from broader group of stakeholders.

Consensus on devrel team was that it was better to use bigint everywhere. I agree they can be annoying, especially when working with numbers, but if everything is a bigint this should really be a problem. It also makes behavior much more predictable knowing most numbers should be bigints

> The test error I get is URLTokenBaseHTTPError: Network request error. Received status 400 (Bad Request): TransactionPool.Remember: txn dead: round 418 outside of 397--407. As far as I can tell a transaction within a group is getting conflicting rounds.

Ah ok I was able to reproduce. Previously I was just running the client test by itself which was passing. When you run the full suite you get tests in parallel which affect the blocks. We just need to provide a wider window for testing.

robdmoore:
Cool, can I make a suggestion that when *receiving* data we accept `number | bigint`, but when relaying information we provide `bigint`. I guess the problem may occur where we have a field that can be both though?

joe-p:
I made the change to just `bigint`. If it becomes a source of a lot of developer problems then we can re-evaluate and add a number union. I think it's easier to add a number union than to remove it. 

I'll be merging this PR and recording bootcamp content with the current state of the base branch.

PR #250, opened 2024-03-19
fix: Include schema in AppClient's create params

### LINKED ISSUE #251: App Client does not allow you to specify extraPages for create calls
## Problem

The `createApp` function allows the specification of storage schema (global and local state + extra pages), but the app client does not allow you to pass this information in. It is taken directly from the AppSpec which means it is not possible to over allocate these values to future proof an application for updates

## Solution

Add an optional schema prop to the create and deploy calls on the app client so these values may be overridden

### Pros and Cons

Pros: Power users can override these values when required
Cons: It would be possible to specify values that aren't valid (ie. allocate less pages than required) and the deployment would fail

tristanmenzel:
## Proposed Changes

- Include schema in appclient's create params

Fixes #251

### DIFF
--- a/src/types/app-client.ts
+++ b/src/types/app-client.ts
@@ -30,6 +30,7 @@ import {
   AppMetadata,
   AppReference,
   AppState,
+  AppStorageSchema,
   BoxName,
   DELETABLE_TEMPLATE_NAME,
   OnSchemaBreak,
@@ -156,7 +157,10 @@ export interface AppClientDeployCallInterfaceParams {
 }
 
 /** Parameters to pass into ApplicationClient.deploy */
-export interface AppClientDeployParams extends AppClientDeployCoreParams, AppClientDeployCallInterfaceParams {}
+export interface AppClientDeployParams extends AppClientDeployCoreParams, AppClientDeployCallInterfaceParams {
+  /** Any overrides for the storage schema to request for the created app; by default the schema indicated by the app spec is used. */
+  schema?: Partial<AppStorageSchema>
+}
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface AppClientCallRawArgs extends RawAppCallArgs {}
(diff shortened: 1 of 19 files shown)


src/types/app-client.ts
```diff
@@ -201,7 +204,12 @@ export type AppClientCreateOnComplete = {
 }
 
 /** Parameters for creating a contract using ApplicationClient */
-export type AppClientCreateParams = AppClientCallParams & AppClientCompilationParams & AppClientCreateOnComplete
+export type AppClientCreateParams = AppClientCallParams &
+  AppClientCompilationParams &
+  AppClientCreateOnComplete & {
+    /** The storage schema to request for the created app */
```
robdmoore, line 210:
```suggestion
    /** Any overrides for the storage schema to request for the created app; by default the schema indicated by the app spec is used. */
```

src/types/app-client.ts
```diff
@@ -201,7 +204,12 @@ export type AppClientCreateOnComplete = {
...
+    schema?: AppStorageSchema
```
robdmoore, line 211:
```suggestion
    schema?: Partial<AppStorageSchema>
```

src/types/app-client.ts
```diff
@@ -156,7 +157,9 @@ export interface AppClientDeployCallInterfaceParams {
 }
 
 /** Parameters to pass into ApplicationClient.deploy */
-export interface AppClientDeployParams extends AppClientDeployCoreParams, AppClientDeployCallInterfaceParams {}
+export interface AppClientDeployParams extends AppClientDeployCoreParams, AppClientDeployCallInterfaceParams {
+  schema?: AppStorageSchema
```
robdmoore, line 161:
```suggestion
  /** Any overrides for the storage schema to request for the created app; by default the schema indicated by the app spec is used. */
  schema?: Partial<AppStorageSchema>
```

tristanmenzel:
@robdmoore  https://github.com/algorandfoundation/algokit-client-generator-ts/pull/104

PR #259, opened 2024-03-25
Added `createAsset` and support 64-bit integers from indexer

robdmoore:
* feat: Added createAsset method
* fix: Updating asset-transfer-transaction.amount and .closeAmount to allow bigint values since they can be 64-bit numbers

BREAKING CHANGE: Indexer's from algokit.getAlgoIndexerClient(...) now have IntDecoding.MIXED rather than IntDecoding.DEFAULT
so that numbers > 53-bit will be accurately represented, you can override this with a new parameter to that function.

Also, a number of fields in types/indexer that can have 64-bit values have been changed from `number` to `number | bigint`.

### DIFF
--- a/src/asset.ts
+++ b/src/asset.ts
@@ -9,7 +9,7 @@ import {
   sendGroupOfTransactions,
   sendTransaction,
 } from './transaction'
-import { AssetBulkOptInOutParams, AssetOptInParams, AssetOptOutParams } from './types/asset'
+import { AssetBulkOptInOutParams, AssetOptInParams, AssetOptOutParams, CreateAssetParams } from './types/asset'
 import { SendTransactionFrom, SendTransactionResult, TransactionGroupToSend, TransactionToSign } from './types/transaction'
 import Algodv2 = algosdk.Algodv2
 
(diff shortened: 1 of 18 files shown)


src/indexer-lookup.spec.ts
```diff
@@ -85,4 +85,71 @@ describe('indexer-lookup', () => {
...
+    await algokit.assetOptIn(
+      {
+        account: secondAccount,
+        assetId: Number(asset.confirmation!.assetIndex),
+      },
+      algod,
+    )
+    const transfer = await algokit.transferAsset(
+      {
+        amount: 134_640_597_783_270_615n,
```
neilcampbell, line 112: Could be nice to extract `134_640_597_783_270_615n` into a constant.

src/asset.ts
```diff
@@ -67,6 +67,77 @@ async function ensureAssetBalanceConditions(
...
+  })
+
+  const encodedLease = encodeLease(lease)
+  if (encodedLease) {
+    transaction.addLease(encodedLease)
+  }
+
+  if (!sendParams.skipSending) {
+    const result = await sendTransaction({ transaction, from: creator, sendParams }, algod)
+    Config.getLogger(sendParams.suppressLog).info(
```
neilcampbell, line 129: Should this be verbose instead?

neilcampbell (commented):
Nice one! A couple of small comments.

PR #261, opened 2024-04-04
Preparing client and composer for release

robdmoore:
This merges with main as well as applying a refactor / full code review of #234

See [15ba017](https://github.com/algorandfoundation/algokit-utils-ts/pull/261/commits/15ba01794a3532c5bfccaf8abeb5e6c44b58e3da) for a clean diff of my refactor.


src/app-deploy.spec.ts
```diff
@@ -36,7 +36,7 @@ describe('deploy-app', () => {
     expect(app.updatable).toBe(creationMetadata.updatable)
     expect(app.deletable).toBe(creationMetadata.deletable)
     expect(app.version).toBe(creationMetadata.version)
-  })
+  }, 20_000)
```
robdmoore, line 40: These overcome the intermittent indexer slowdown that exists on current localnet

src/index.ts
```diff
@@ -12,5 +12,3 @@ export * from './localnet'
 export * from './network-client'
 export * from './transaction'
 export * from './transfer'
-export * from './types/algorand-client'
```
robdmoore, line 15: While these classes are in beta it makes sense to not expose them from index so you have to explicitly import them

src/types/algorand-client.spec.ts
```diff
@@ -98,7 +99,7 @@ describe('AlgorandClient', () => {
       sender: alice.addr,
       appId: appId,
       method: appClient.appClient.getABIMethod('txnArg')!,
-      args: [{ type: 'pay' as const, sender: alice.addr, receiver: alice.addr, amount: algokit.microAlgos(0) }],
+      args: [algorand.transactions.payment({ sender: alice.addr, receiver: alice.addr, amount: algokit.microAlgos(0) })],
```
robdmoore, line 102: By allowing `Promise<Transaction>` we get to have a similarly nice syntax, but without leaking the internal `Txn` type.

src/types/algorand-client.spec.ts
```diff
@@ -111,20 +112,19 @@ describe('AlgorandClient', () => {
...
 
   test('method with method call arg', async () => {
-    const helloWorldParams = {
-      type: 'methodCall' as const,
+    const helloWorldCall = {
       sender: alice.addr,
       appId: appId,
       method: appClient.appClient.getABIMethod('helloWorld')!,
-    }
+    } satisfies MethodCallParams
```
robdmoore, line 119: You don't /have/ to use `satisfies MethodCallParams`, but by doing so you'd get more obvious intellisense.

src/types/algorand-client.ts
```diff
@@ -1,36 +1,21 @@
...
-  OnlineKeyRegParams,
-  PayTxnParams,
-} from './composer'
+import AlgokitComposer, { ExecuteParams, MethodCallParams } from './composer'
 import { AlgoConfig } from './network-client'
-import { SendAtomicTransactionComposerResults, SendTransactionFrom } from './transaction'
+import { ConfirmedTransactionResult, SendAtomicTransactionComposerResults, SendTransactionFrom } from './transaction'
+import Transaction = algosdk.Transaction
 
-async function unwrapSingleSendResult(results: Promise<SendAtomicTransactionComposerResults>) {
```
robdmoore, line 21: I collapsed this method into the class since it was only used in one place and I was able to achieve a more terse syntax.

src/types/composer.ts
```diff
@@ -1,15 +1,36 @@
...
+export type RoundNumberParam = bigint
+
+/** Transaction parameter that provides an asset amount. */
+export type AssetAmountParam = bigint
+
+/** Transaction parameter that provides an asset ID. */
+export type AssetIdParam = bigint
+
+/** Transaction parameter that provides an application ID. */
+export type AppIdParam = bigint
```
robdmoore, line 24: Adding explicit types for the kind of information being asked for - makes intellisense nicer / more obvious and allows us to see exactly what types we are accepting for each bit of information.

src/types/composer.ts
```diff
@@ -155,41 +185,57 @@ export type AppCallParams = CommonTxnParams & {
...
+     * * A transaction (where the signer will be automatically assigned)
+     * * An unawaited transaction (e.g. from algorand.transactions.transactionType())
+     * * Another method call (via method call params object)
+     */
+    args?: (algosdk.ABIValue | TransactionWithSigner | Transaction | Promise<Transaction> | MethodCallParams)[]
   }
 
 type Txn =
-  | (PayTxnParams & { type: 'pay' })
+  | (PaymentParams & { type: 'pay' })
```
robdmoore, line 217: Removing abbreviation for more clarity.

src/types/composer.ts
```diff
@@ -201,101 +247,164 @@ export type AlgokitComposerParams = {
...
+ */
 export default class AlgokitComposer {
   /** Map of txid to ABI method */
   private txnMethodMap: Map<string, algosdk.ABIMethod> = new Map()
 
   /** Transactions that have not yet been composed */
   private txns: Txn[] = []
 
-  /** The underlying AtomicTransactionComposer */
-  atc: algosdk.AtomicTransactionComposer
```
robdmoore, line 212: Turns out we didn't need this as state and by creating it on the fly we can call `buildGroup` multiple times without error.

src/types/composer.ts
```diff
@@ -310,10 +419,10 @@ export default class AlgokitComposer {
     return txnWithSigners
   }
 
-  private commonTxnBuildStep(params: CommonTxnParams, txn: algosdk.Transaction, suggestedParams: algosdk.SuggestedParams) {
-    if (params.lease) new Uint8Array(Buffer.from(params.lease))
+  private commonTxnBuildStep(params: CommonTransactionParams, txn: algosdk.Transaction, suggestedParams: algosdk.SuggestedParams) {
+    if (params.lease) txn.addLease(encodeLease(params.lease)!)
```
robdmoore, line 423: This wasn't setting the lease :P, also now using the nice encodeLease helper from utils

src/types/composer.ts
```diff
@@ -335,6 +444,7 @@ export default class AlgokitComposer {
       txn.fee = txn.estimateSize() * suggestedParams.fee || algosdk.ALGORAND_MIN_TX_FEE
       if (params.extraFee) txn.fee += params.extraFee.microAlgos
     }
+    txn.flatFee = true
```
robdmoore, line 447: I assume this is needed unless I'm misunderstanding the fee control?

src/types/composer.ts
```diff
@@ -343,80 +453,53 @@ export default class AlgokitComposer {
...
     const methodAtc = new algosdk.AtomicTransactionComposer()
 
     methodAtc.addMethodCall({
       ...params,
       appID: Number(params.appId || 0),
-      note: params.note ? new Uint8Array(Buffer.from(params.note)) : undefined,
-      lease: params.lease ? new Uint8Array(Buffer.from(params.lease)) : undefined,
+      note: encodeTransactionNote(params.note),
+      lease: encodeLease(params.lease),
+      rekeyTo: params.rekeyTo,
```
robdmoore, line 502: I realised we don't have fee control here - is that a problem?

src/types/composer.ts
```diff
@@ -442,49 +525,17 @@ export default class AlgokitComposer {
...
-  }
-
-  private buildAppCall(params: AppCallParams, suggestedParams: algosdk.SuggestedParams) {
-    const sdkParams = {
-      from: params.sender,
+      assetMetadataHash: params.metadataHash,
+      manager: params.manager,
+      reserve: params.reserve,
+      freeze: params.freeze,
+      clawback: params.clawback,
```
robdmoore, line 536: This stuff was missing from create asset

src/types/composer.ts
```diff
@@ -636,30 +733,42 @@ export default class AlgokitComposer {
...
+  async execute(params?: ExecuteParams): Promise<SendAtomicTransactionComposerResults> {
+    const { atc, transactions: group } = await this.build()
 
     let waitRounds = params?.maxRoundsToWaitForConfirmation
-
     if (waitRounds === undefined) {
       const lastRound = group.reduce((max, txn) => Math.max(txn.txn.lastRound, max), 0)
       const { firstRound } = await this.getSuggestedParams()
-      waitRounds = lastRound - firstRound
+      waitRounds = lastRound - firstRound + 1
```
robdmoore, line 765: I figured we probably need to leave 1 extra round as leeway?

src/types/composer.ts
joe-p: I had this as a property in the event of calling `buildGroup` prior to `execute`. It's important to preserve the same ATC between the two calls otherwise the suggestedParams may change. I realize this means we should probably expose the ATC status and add a check in execute

src/types/composer.ts
```diff
@@ -155,41 +185,57 @@ export type AppCallParams = CommonTxnParams & {
...
-    /** Arguments to the ABI method */
-    args?: (algosdk.ABIValue | Txn)[]
+    /** Arguments to the ABI method, either:
+     * * An ABI value
+     * * A transaction with explcit signer
+     * * A transaction (where the signer will be automatically assigned)
+     * * An unawaited transaction (e.g. from algorand.transactions.transactionType())
+     * * Another method call (via method call params object)
+     */
+    args?: (algosdk.ABIValue | TransactionWithSigner | Transaction | Promise<Transaction> | MethodCallParams)[]
```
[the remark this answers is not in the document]
joe-p: I like it. Leaves `MethodCallParams` as a bit of an odd-man out but nested method calls are a more advanced feature anyways so probably fine as is.

src/types/composer.ts
joe-p: Ah yep good call

src/types/composer.ts
joe-p: I feel the other way around. `string` makes for nicer Intellisense than `AccountAddressParam`. I expect to see the type here, not a description of the value. One can read the TSDoc description to understand what the value should be.

src/types/composer.ts
joe-p: And if you want to build multiple times, I'd propose a `rebuild` method

src/index.ts
joe-p: As discussed on Slack, will add this back to index but preserve the beta tag

src/types/composer.ts
robdmoore: Feel free to try and refactor it back, I pulled it out to solve for a case where I needed to call buildGroup preemptively and then it caused the underlying atc to error

src/types/composer.ts
joe-p: Removed for simplicity in [560195d](https://github.com/algorandfoundation/algokit-utils-ts/pull/261/commits/560195d8b12e03a6045444e24f891b61609b5e1a)

src/types/composer.ts
joe-p: [61f0428](https://github.com/algorandfoundation/algokit-utils-ts/pull/261/commits/61f04288fbe820105bf89a69cf0d43211a47884e) should prevent errors when calling build multiple times while also adding a `rebuild` method for explicitly rebuilding with new sp

src/types/composer.ts
joe-p: Added in [c16635b](https://github.com/algorandfoundation/algokit-utils-ts/pull/261/commits/c16635b32ae760344a7800ac81461f395821a543)

src/index.ts
joe-p: Exported in [6cf626b](https://github.com/algorandfoundation/algokit-utils-ts/pull/261/commits/6cf626b83c2c5565a7f9bbfb22421b577f2ee67b)

PR #278, opened 2024-05-20
Fix/resource_pop_bugs

joe-p:
## Proposed Changes

Fixes two bugs on resource population

1. Pushing a resource to an undefined array doesn't actually work

2. Box references weren't checking if the app was available

### DIFF
--- a/src/transaction/transaction.ts
+++ b/src/transaction/transaction.ts
@@ -427,6 +427,14 @@ export async function populateAppCallResources(atc: algosdk.AtomicTransactionCom
     } else if (type === 'box') {
       const { app, name } = reference as algosdk.modelsv2.BoxReference
       txns[txnIndex].txn.boxes = [...(txns[txnIndex].txn.boxes ?? []), { appIndex: Number(app), name }]
+
+      // Add the app if it is not already available
+      let appAlreadyAvailable = false
+      txns.forEach((t) => {
+        if (t.txn.appIndex === Number(reference)) appAlreadyAvailable = true
+        if (t.txn.appForeignApps?.includes(Number(reference))) appAlreadyAvailable = true
+      })
+      if (!appAlreadyAvailable) populateGroupResource(txns, app, 'app')
     } else if (type === 'assetHolding') {
       const { asset, account } = reference as algosdk.modelsv2.AssetHoldingReference
       txns[txnIndex].txn.appForeignAssets = [...(txns[txnIndex].txn.appForeignAssets ?? []), Number(asset)]
@@ -468,6 +476,9 @@ export async function populateAppCallResources(atc: algosdk.AtomicTransactionCom
 
     g.boxes?.forEach((b) => {
       populateGroupResource(group, b, 'box')
+
+      // Remove resources from the group if we're adding them here
+      g.apps = g.apps?.filter((app) => BigInt(app) !== BigInt(b.app))
     })
 
     g.assets?.forEach((a) => {
--- a/src/transaction/transaction.ts
+++ b/src/transaction/transaction.ts
@@ -298,15 +298,15 @@ export async function populateAppCallResources(atc: algosdk.AtomicTransactionCom
 
     // Do accounts first because the account limit is 4
     r.accounts?.forEach((a) => {
-      group[i].txn.appAccounts?.push(algosdk.decodeAddress(a))
+      group[i].txn.appAccounts = [...(group[i].txn.appAccounts ?? []), algosdk.decodeAddress(a)]
     })
 
     r.apps?.forEach((a) => {
-      group[i].txn.appForeignApps?.push(Number(a))
+      group[i].txn.appForeignApps = [...(group[i].txn.appForeignApps ?? []), Number(a)]
     })
 
     r.assets?.forEach((a) => {
-      group[i].txn.appForeignAssets?.push(Number(a))
+      group[i].txn.appForeignAssets = [...(group[i].txn.appForeignAssets ?? []), Number(a)]
     })
 
     const accounts = group[i].txn.appAccounts?.length || 0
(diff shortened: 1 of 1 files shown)


joe-p:
@robdmoore We're seeing the indexer tests timeout and I'm also seeing the same locally since I've updated my localnet. I know we've talked about this before but forget what the reasoning was and how we might reasonably fix this

src/transaction/transaction.ts
```diff
@@ -421,22 +421,32 @@ export async function populateAppCallResources(atc: algosdk.AtomicTransactionCom
...
       const { app, name } = reference as algosdk.modelsv2.BoxReference
-      txns[txnIndex].txn.boxes?.push({ appIndex: Number(app), name: name })
+      txns[txnIndex].txn.boxes = [...(txns[txnIndex].txn.boxes ?? []), { appIndex: Number(app), name }]
+
+      if (app !== 0) {
+        // Add the app if it is not already available
+        let appAlreadyAvailable = false
+        txns.forEach((t) => {
+          if (t.txn.appIndex === Number(reference)) appAlreadyAvailable = true
+          if (t.txn.appForeignApps?.includes(Number(reference))) appAlreadyAvailable = true
```
kylebeee, line 436: probably clearer to do a single if with an `||` instead of two if statements & two assignments

kylebeee (approved):
Looks good, keep in mind i'm missing most of the context of these changes. The changes themselves look good though

src/transaction/transaction.ts
joe-p: Yeah true. I simplified the assignment and switched to `for...of` so we can break the loop in 75498cb696cc72ea9ded3e992138884801645c46

kylebeee: github needs to add the fire emoji

src/transaction/transaction.ts
```diff
@@ -298,15 +298,15 @@ export async function populateAppCallResources(atc: algosdk.AtomicTransactionCom
 
     // Do accounts first because the account limit is 4
     r.accounts?.forEach((a) => {
-      group[i].txn.appAccounts?.push(algosdk.decodeAddress(a))
+      group[i].txn.appAccounts = [...(group[i].txn.appAccounts ?? []), algosdk.decodeAddress(a)]
```
neilcampbell, line 301:
```suggestion
      group[i].txn.appAccounts = (group[i].txn.appAccounts ?? []).concat(algosdk.decodeAddress(a))
```
Alternatively you can use concat.

robdmoore (approved):

src/transaction/transaction.ts
```diff
@@ -421,22 +421,32 @@ export async function populateAppCallResources(atc: algosdk.AtomicTransactionCom
...
+      txns[txnIndex].txn.appAccounts = [...(txns[txnIndex].txn.appAccounts ?? []), algosdk.decodeAddress(reference as string)]
     } else if (type === 'app') {
-      txns[txnIndex].txn.appForeignApps?.push(Number(reference))
+      txns[txnIndex].txn.appForeignApps = [...(txns[txnIndex].txn.appForeignApps ?? []), Number(reference)]
     } else if (type === 'box') {
       const { app, name } = reference as algosdk.modelsv2.BoxReference
-      txns[txnIndex].txn.boxes?.push({ appIndex: Number(app), name: name })
+      txns[txnIndex].txn.boxes = [...(txns[txnIndex].txn.boxes ?? []), { appIndex: Number(app), name }]
+
+      if (app !== 0) {
```
neilcampbell, line 431: This condition won't work for BigInt(0).

neilcampbell (approved):

src/transaction/transaction.spec.ts
```diff
@@ -327,7 +327,7 @@ describe('transaction', () => {
...
+        }, data {AccountBaseData:{Status:Offline MicroAlgos:{Raw:9998000} RewardsBase:${rewardsBase} RewardedMicroAlgos:{Raw:0} AuthAddr:AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY5HFKQ IncentiveEligible:false TotalAppSchema:{_struct:{} NumUint:0 NumByteSlice:0} TotalExtraAppPages:0 TotalAppParams:0 TotalAppLocalStates:0 TotalAssetParams:0 TotalAssets:0 TotalBoxes:0 TotalBoxBytes:0 LastProposed:0 LastHeartbeat:0} VotingData:{VoteID:[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] SelectionID:[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] StateProofID:[0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] VoteFirstValid:0 VoteLastValid:0 VoteKeyDilution:0}}, tried to spend {9999999999999})`,
```
neilcampbell, line 330: FYI, I've temporarily pinned the algod docker image version used in the latest algokit. The build agent, will use this and therefore you can revert the change to this file.

robdmoore:
@joe-p https://github.com/algorand/indexer/issues/1613

src/transaction/transaction.ts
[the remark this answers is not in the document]
joe-p: `.push` returns the array length rather than the array itself

src/transaction/transaction.ts
joe-p: Ah good catch. Used toString in [41bbfbe](https://github.com/algorandfoundation/algokit-utils-ts/pull/278/commits/41bbfbe727b1cabe53a45209b789985c5b9692c7)

joe-p:
> There's some quite complex logic here that it may make sense to have test coverage for, but on the other hand I know this logic will move to algod at some point so probably time better spent adding test coverage there when that happens?

Yeah that's exactly what I'm thinking. Will start algod effort this week so didn't want to spend time writing the tests here. I've confirmed that this works in the reti ui.

src/transaction/transaction.ts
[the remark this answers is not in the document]
robdmoore: Good point, I probably meant `concat` then haha

joe-p: Ah forgot about concat. Some quick google-fu tells me there isn't a significant tradeoff here so just keeping as is.

neilcampbell: @joe-p Yeah nothing massive in this context. There is when you use the spread operator with `add` for example https://jsfiddle.net/y0fjxLv5/.

PR #279, opened 2024-05-21
fix: patch to propagate upon algokit 2.0.6 release (removal of ConsensusProtocol setting from network config)

aorumbayev:
## Proposed Changes

- To be merged before changes in https://github.com/algorandfoundation/algokit-cli/pull/505 are released


robdmoore (approved):

joe-p:
Is there a reason why we're checking against the full error message? Would be more flexible if we just match a regex rather than the full string.

neilcampbell:
@joe-p Yeah agreed, it's probably a little strict. I'll adjust.

neilcampbell (approved):

PR #284, opened 2024-05-28
feat: fix signers during resource population

joe-p:
## Proposed Changes

- The initial plan was to wait for upstream algod fix, but multiple users have requested this feature now.  The only downside is increased algod calls, but since resource population is disabled by default I think it is acceptable. 

Won't be long until https://github.com/algorand/go-algorand/pull/5942 is merged and we can remove this code and use the upstream fix.

### DIFF
--- a/src/transaction/transaction.spec.ts
+++ b/src/transaction/transaction.spec.ts
@@ -783,4 +783,36 @@ describe('Resource Packer: meta', () => {
 
     expect(res.transaction.appAccounts?.length || 0).toBe(0)
   })
+
+  test('rekeyed account', async () => {
+    const { testAccount } = fixture.context
+    const { algorand } = fixture
+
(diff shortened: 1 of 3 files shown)


src/transaction/transaction.spec.ts
```diff
@@ -783,4 +783,36 @@ describe('Resource Packer: meta', () => {
...
     expect(res.transaction.appAccounts?.length || 0).toBe(0)
   })
+
+  test('rekeyed account', async () => {
+    const { testAccount } = fixture.context
+    const { algorand } = fixture
+
+    const authAddr = algorand.account.random().account
+
+    await algorand.send.payment
```
neilcampbell, line 793:
```suggestion
```

neilcampbell (approved):
Just pushed a doc update commit, so the build passes.

PR #287, opened 2024-06-09
feat: Added initial documentation and tweaks to deprecate account, network-client and localnet functions

robdmoore:
Inspired by #286, but a slightly different approach that starts to lay the foundation for deprecating the old way of using AlgoKit Utils in favour of the new approach using AlgorandClient.

This method allows us to keep this library in a releasable state while working on the deprecation bit by bit.

The patterns in this pull request help us guide users towards where we want and do proper deprecation of old vs new world, keep the documentation clear, keep the migration path clear and not have to do the migration all in a big bang.

Deprecates the following in favour of new `AlgorandClient` exposed functionality:
* `algokit.multisigAccount`
* `algokit.rekeyedAccount`
* `algokit.transactionSignerAccount`
* `algokit.randomAccount`
* `algokit.mnemonicAccountFromEnvironment`
* `algokit.getAccountAddressAsUint8Array`
* `algokit.getAccountAddressAsString`
* `algokit.getAccountInformation`
* `algokit.getAccountAssetInformation`
* `algokit.getAccount`
* `algokit.getDispenserAccount`
* `algokit.mnemonicAccount`
* `algokit.getTestNetDispenserApiClient`
* `algokit.getKmdWalletAccount`
* `algokit.getKmdWalletAccount`
* `algokit.getOrCreateKmdWalletAccount`
* `algokit.isLocalNet`
* `algokit.getConfigFromEnvOrDefaults`
* `algokit.getAlgodConfigFromEnvironment`
* `algokit.getIndexerConfigFromEnvironment`
* `algokit.getAlgoNodeConfig`
* `algokit.getDefaultLocalNetConfig`
* `algokit.getAlgoClient`
* `algokit.getAlgoIndexerClient`
* `algokit.getAlgoKmdClient`
* `algokit.isTestNet`
* `algokit.isMainNet`
* `algorandFixture(undefined, config)` override (now you can pass in clients or client configs to the first parameter of `algorandFixture`)
* `AccountConfig` in `types/account`

Changes:
* `getTestAccount` to take an `AlgorandClient` rather than `algod` and `kmd` (no change for `getTestAccount` method exposed from `algorandFixture` though, which is the primary way of accessing it (outside of a tweak to the recently added `accountGetter` parameter to also take an `AlgorandClient`)
* `algorandFixture().context` now has an `algorand` property with an `AlgorandClient` instance
* `dispenser()` method in `AccountManager` is now called `dispenserFromEnvironment()` to reflect the fact it loads environment variables and expects a Node.js environment

Additions:
* `getAccountTransactionSigner` method exposed from `types/account-manager` (similar to previous `getSenderTransactionSigner` method)
* `AccountManager` has a `kmd` property with an instance of the new `KmdAccountManager` class, and a `rekeyed` method that tracks a `TransactionSignerAccount` to a different sender
* `KmdAccountManager` class with `getWalletAccount`, `getOrCreateWalletAccount` and `getLocalNetDispenserAccount` methods
* `ClientManager` has new instance methods for `network()`, `isLocalNet()`, `isTestNet()` and `isMainNet()` and static methods for `getConfigFromEnvironmentOrLocalNet`, `getAlgodConfigFromEnvironment`, `getIndexerConfigFromEnvironment`, `getAlgoNodeConfig`, `getDefaultLocalNetConfig`, `getAlgodClient`, `getAlgodClientFromEnvironment`, `getIndexerClient`, `getIndexerClientFromEnvironment`, `getKmdClient`, and `getKmdClientFromEnvironment`


docs/README.md
```diff
@@ -33,13 +33,47 @@ This library requires `algosdk` as a peer dependency. Ensure you have it install
...
 
-Then you can use intellisense to auto-complete the various functions that are available by typing `algokit.` in your favourite Integrated Development Environment (IDE), or you can refer to the [reference documentation](code/modules/index.md).
+> [!WARNING]
+> Previous versions of AlgoKit Utils encouraged you to include an import that looks like this (note the subtle difference of the extra `* as`):
+>
+> ```typescript
+> import * as algokit from '@algorandfoundation/algokit-utils'
+> ```
+>
+> This version will still work, but it exposes an older, function-based interface to the functionality that is in the process of being deprecated. The recommended way to use AlgoKit Utils is via the `AlgorandClient` class mentioned below, which is easier and more convenient to use. Some functionality won't yet be migrated to the new approach and this old approach will be needed, but the documentation pages will indicate when this is the case.
```
robdmoore, line 46: This allows us to a) make a clean break between legacy and new way of using AlgoKit Utils without a breaking change. It also allows us to link to this Usage section across the docs in any part of the docs that is referencing the legacy stuff so we don't have to replace everything in a big bang. We can do a few pull requests one-by-one each each with a methodical focus so we are less likely to get stuff wrong. The focus of this PR is deprecating account, network-client and localnet.

docs/capabilities/account.md
```diff
@@ -1,10 +1,31 @@
...
+const accountManager = new AccountManager(clientManager)
+```
+
+## `TransactionSignerAccount`
+
+The core internal type that holds information about a signer/sender pair for a transaction is [`TransactionSignerAccount`](../code/interfaces/types_account.TransactionSignerAccount.md), which represents an `algosdk.TransactionSigner` (`signer`) along with a sender address (`addr`) as the encoded string address.
+
+Many methods in `AccountManager` expose a `TransactionSignerAccount`.
+
+### `SendTransactionFrom`
```
robdmoore, line 23: A lot of the existing docs link to this jump link so important to keep it for now, added the legacy note.

docs/capabilities/algokit-composer.md
```diff
@@ -0,0 +1,3 @@
+# AlgoKit composer
+
+TODO
```
robdmoore, line 3: This can be tackled in future pull requests, it makes sense to have a transaction focussed one that also takes care of adding this.

docs/capabilities/algorand-client.md
```diff
@@ -0,0 +1,38 @@
+# Algorand client
```
robdmoore, line 1: This page needs more stuff, but this is a good starting point covering the stuff relevant for this pull request

docs/capabilities/app-client.md
```diff
@@ -2,17 +2,52 @@
 
 Application client that works with ARC-0032 application spec defined smart contracts (e.g. via Beaker).
 
+> [!NOTE]
+> This page covers the untyped app client, but it's worth exploring the [typed client generation feature in AlgoKit](https://github.com/algorandfoundation/algokit-cli/blob/main/docs/features/generate.md#1-typed-clients), which will give you a better developer experience with strong typing and intellisense specific to the app itself.
+
```
robdmoore, line 7: This is actually a huge gap in the AlgoKit Utils docs. I think it's important in a future pull requests that focusses on apps that we pull in typed app client docs - they are currently spread between the generator repo and the algokit cli repo.

docs/capabilities/client.md
```diff
@@ -4,11 +4,30 @@ Client management is one of the core capabilities provided by AlgoKit Utils. It
```
robdmoore, line 4: This page now covers `ClientManager` pretty comprehensively.

src/account/account.spec.ts
```diff
@@ -1,65 +0,0 @@
-import { test } from '@jest/globals'
```
robdmoore, line 1: This got moved to types/account-manager.spec.ts

src/account/account.ts
```diff
@@ -1,13 +1,10 @@
 import algosdk from 'algosdk'
```
robdmoore, line 1: Every function in this (account) folder is now marked deprecated and the implementation points to the new place.

src/types/account-manager.ts
```diff
@@ -257,22 +388,33 @@ export class AccountManager {
...
    * @returns The account
    */
-  public async dispenser() {
-    return this.signerAccount(await getDispenserAccount(this._clientManager.algod, this._clientManager.kmd))
+  public async dispenserFromEnvironment() {
+    if (!process || !process.env) {
+      throw new Error('Attempt to get dispenser from environment from a non Node.js context; this is not supported!')
+    }
+
+    return process.env[`${DISPENSER_ACCOUNT.toUpperCase()}_MNEMONIC`]
```
robdmoore, line 415: This is technically a breaking change - previously we would get LocalNet if pointing to LocalNet, but I think it's much better to use environment variables in preference, not only is that less HTTP calls (no isLocalNet check), but it means that you can control the dispenser to be a different account if you want. I think the chances of this breaking someone in reality would be slim.

src/types/account-manager.ts
```diff
@@ -257,22 +388,33 @@ export class AccountManager {
...
    * ```typescript
-   * const account = await account.dispenser()
+   * const account = await account.dispenserFromEnvironment()
    * ```
-   * If running on LocalNet then it will return the default dispenser account automatically,
-   *  otherwise it will load the account mnemonic stored in process.env.DISPENSER_MNEMONIC.
+   *
    * @returns The account
    */
-  public async dispenser() {
```
robdmoore, line 274: This is a breaking naming change. Previously it was called `dispenser`, but now it's called `dispenserFromEnvironment`, which matches the convention that anything that pulls from environment variables now has "FromEnvironment" in the name. These methods all need Node.js, so having a clear demarcation of this is important I think.

Given this class is part of the new AlgorandClient ecosystem that is marked as beta, it's expected there may be some breaking changes so I think it's OK to do this without bumping major version, but happy if anyone disagrees with this.

src/index.ts
```diff
@@ -1,15 +1,31 @@
-export * from './account'
-export * from './amount'
+// TODO review exports
```
robdmoore, line 1: Tracking deprecation progress via these comments

src/index.ts
```diff
@@ -1,15 +1,31 @@
...
+// Default export with intellisense of the most pertinent parts of AlgoKit Utils
+import * as amount from './amount'
+import * as config from './config'
+import * as algorand from './types/algorand-client'
+
+export default {
+  ...amount,
+  ...config,
+  AlgorandClient: algorand.AlgorandClient,
+}
```
robdmoore, line 31: Would be great to get feedback about this. I think it's a good idea to make a clean break against the default way you import algokit utils and get baseline intellisense. This should end up being a very small set of things since most things will come from AlgorandClient.

It results in a cleaner import of `import algokit from "package"` rather than `import * as algokit from "package"` and means you can quickly tell with a project wide search if you are using new or old way of using algokit utils across a project and gradually migrate.

src/indexer-lookup.spec.ts
```diff
@@ -29,16 +29,16 @@ describe('indexer-lookup', () => {
   }, 20_000)
 
   test('Account is found by id', async () => {
-    const { indexer, testAccount, transactionLogger } = localnet.context
-    await transactionLogger.waitForIndexer(indexer)
+    const { indexer, testAccount } = localnet.context
+    await runWhenIndexerCaughtUp(() => algokit.lookupAccountByAddress(testAccount.addr, indexer))
```
robdmoore, line 33: Not sure why I needed this, but didn't previously. Could mean I'd done something wrong somewhere, or just be part of the evolving mess of indexer performance. I was getting a consistent 404 for this test when running it along with other tests (but not when running it solo). Adding this meant it started working.

src/localnet/get-kmd-wallet-account.ts
```diff
@@ -1,11 +1,13 @@
 import algosdk from 'algosdk'
```
robdmoore, line 1: Everything in this folder (localnet) is now deprecated

src/network-client.ts
```diff
@@ -1,106 +1,65 @@
 import algosdk from 'algosdk'
```
robdmoore, line 1: These are all deprecated now

src/testing/account.ts
```diff
@@ -1,37 +1,35 @@
...
  * @param param0 The config for the test account to generate
- * @param algod An algod client
- * @param kmd A KMD client, if not specified then a default KMD client will be loaded from environment variables
+ * @param algorand An AlgorandClient client
  * @returns The account, with private key loaded
  */
 export async function getTestAccount(
   { suppressLog, initialFunds, accountGetter }: GetTestAccountParams,
-  algod: Algodv2,
-  kmd?: Kmd,
```
robdmoore, line 22: Technically this is a breaking change, but I'd be very surprised if anyone used this method directly - you have to go out of your way to import it (it's not in the default `* as algokit` export. The way it's exposed via `algorandFixture` isn't breaking.

src/testing/fixtures/algorand-fixture.ts
```diff
@@ -86,34 +72,36 @@ export function algorandFixture(fixtureConfig: AlgorandFixtureConfig | undefined
...
  * @returns The fixture
  */
+export function algorandFixture(fixtureConfig: AlgorandFixtureConfig | undefined, config: AlgoConfig): AlgorandFixture
+
 export function algorandFixture(fixtureConfig?: AlgorandFixtureConfig, config?: AlgoConfig): AlgorandFixture {
-  config = config || getConfigFromEnvOrDefaults()
+  fixtureConfig = { ...fixtureConfig, ...config }
+  if (!fixtureConfig.algod || !fixtureConfig.indexer || !fixtureConfig.kmd) {
+    fixtureConfig = { ...ClientManager.getConfigFromEnvironmentOrLocalNet(), ...fixtureConfig }
+  }
```
robdmoore, line 81: I've changed the signature to allow config OR clients to be passed in via `AlgorandFixtureConfig` type now, much cleaner than a weird way of calling it where you add `undefined` to the first parameter and this code here is less confusing now too. I've marked the override as deprecated to reflect this change, that version will still work so it's not a breaking change.

src/transaction/transaction.ts
```diff
@@ -143,6 +134,15 @@ export const getTransactionWithSigner = async (
         }
 }
 
+const memoize = <T = unknown, R = unknown>(fn: (val: T) => R) => {
```
robdmoore, line 137: Just moving it next to where it's used

src/testing/fixtures/algorand-fixture.ts
```diff
@@ -86,34 +72,36 @@ export function algorandFixture(fixtureConfig: AlgorandFixtureConfig | undefined
...
-      { initialFunds: fixtureConfig?.testAccountFunding ?? algos(10), suppressLog: true },
-      transactionLoggerAlgod,
-      kmd,
-    )
     algorandClient = algorandClient ?? AlgorandClient.fromClients({ algod: transactionLoggerAlgod, indexer, kmd })
+    const acc = await getTestAccount({ initialFunds: fixtureConfig?.testAccountFunding ?? algos(10), suppressLog: true }, algorandClient)
     algorandClient.setSignerFromAccount(acc).setDefaultValidityWindow(1000).setSuggestedParamsTimeout(0)
     const testAccount = { ...acc, signer: algorandClient.account.getSigner(acc.addr) }
     context = {
+      algorand: algorandClient,
```
robdmoore, line 98: Added this easier for better ergonomics if grabbing other stuff from the context.

src/types/account-manager.ts
```diff
@@ -1,35 +1,61 @@
...
+ */
+export const getAccountTransactionSigner = memoize(function (
+  account: TransactionSignerAccount | Account | SigningAccount | LogicSigAccount | MultisigAccount,
+): TransactionSigner {
+  return 'signer' in account
+    ? account.signer
+    : 'lsig' in account
+      ? algosdk.makeLogicSigAccountTransactionSigner(account)
+      : algosdk.makeBasicAccountTransactionSigner(account)
+})
```
robdmoore, line 34: This is copied from `getSenderTransactionSigner` function, but `SendTransactionFrom` is removed in lieu of specifying the types of account classes supported explicitly. This allows us to eventually deprecate it and means this file doesn't take any dependencies on legacy methods anymore.

src/types/account-manager.ts
```diff
@@ -46,33 +80,52 @@ export class AccountManager {
...
    */
-  private signerAccount<T extends SendTransactionFrom>(account: T): TransactionSignerAccount & { account: T } {
+  private signerAccount<T extends TransactionSignerAccount | Account | SigningAccount | LogicSigAccount | MultisigAccount>(
+    account: T,
+  ): TransactionSignerAccount & { account: T } {
     const acc = {
-      addr: getSenderAddress(account),
-      signer: getSenderTransactionSigner(account),
+      addr: 'addr' in account ? account.addr : account.address(),
+      signer: getAccountTransactionSigner(account),
```
robdmoore, line 91: Removing dependencies on legacy methods

src/types/account-manager.ts
```diff
@@ -46,33 +80,52 @@ export class AccountManager {
...
+   *  .setSignerFromAccount(algosdk.generateAccount())
+   *  .setSignerFromAccount(new algosdk.LogicSigAccount(program, args))
+   *  .setSignerFromAccount(new SigningAccount(mnemonic, sender))
+   *  .setSignerFromAccount(new MultisigAccount({version: 1, threshold: 1, addrs: ["ADDRESS1...", "ADDRESS2..."]}, [account1, account2]))
+   *  .setSignerFromAccount({addr: "SENDERADDRESS", signer: transactionSigner})
+   * ```
    * @returns The `AccountManager` instance for method chaining
    */
-  public setSignerFromAccount(account: TransactionSignerAccount | SendTransactionFrom) {
+  public setSignerFromAccount(account: TransactionSignerAccount | Account | LogicSigAccount | SigningAccount | MultisigAccount) {
```
robdmoore, line 115: Removing dependency on legacy type

src/types/account-manager.ts
```diff
@@ -119,7 +187,28 @@ export class AccountManager {
    * @returns The account information
    */
   public async getInformation(sender: string | TransactionSignerAccount): Promise<AccountInformation> {
-    return getAccountInformation(sender, this._clientManager.algod)
+    const account = AccountInformationModel.from_obj_for_encoding(
```
robdmoore, line 190: Deprecating legacy function

src/types/account-manager.ts
```diff
@@ -155,8 +253,24 @@ export class AccountManager {
    * @returns The account
    */
   public fromMnemonic(mnemonicSecret: string, sender?: string) {
-    const account = mnemonicAccount(mnemonicSecret)
-    return this.signerAccount(sender ? rekeyedAccount(account, sender) : account)
+    const account = algosdk.mnemonicToSecretKey(mnemonicSecret)
+    return this.signerAccount(new SigningAccount(account, sender))
```
robdmoore, line 257: Deprecating legacy functions

src/types/account-manager.ts
```diff
@@ -155,8 +253,24 @@ export class AccountManager {
...
+   * @example
+   * ```typescript
+   * const account = await account.fromMnemonic("mnemonic secret ...")
+   * const rekeyedAccount = await account.rekeyed(account, "SENDERADDRESS...")
+   * ```
+   * @param account The account to use as the signer for this new rekeyed account
+   * @param sender The sender address to use as the new sender
+   * @returns The account
+   */
+  public rekeyed(account: TransactionSignerAccount, sender: string) {
```
robdmoore, line 272: Missing convenience method that we previously had

src/types/account-manager.ts
```diff
@@ -187,7 +301,24 @@ export class AccountManager {
...
+      const signer = algosdk.mnemonicToSecretKey(accountMnemonic)
+      return this.signerAccount(new SigningAccount(signer, sender))
+    }
+
+    if (await this._clientManager.isLocalNet()) {
+      const account = await this._kmdAccountManager.getOrCreateWalletAccount(name, fundWith)
+      return this.signerAccount(account.account)
+    }
+
+    throw new Error(`Missing environment variable ${name.toUpperCase()}_MNEMONIC when looking for account ${name}`)
```
robdmoore, line 321: Deprecating legacy method

src/types/account-manager.ts
```diff
@@ -229,7 +360,7 @@ export class AccountManager {
    * @returns A multisig account wrapper
    */
   public multisig(multisigParams: algosdk.MultisigMetadata, signingAccounts: (algosdk.Account | SigningAccount)[]) {
-    return this.signerAccount(multisigAccount(multisigParams, signingAccounts))
+    return this.signerAccount(new MultisigAccount(multisigParams, signingAccounts))
```
robdmoore, line 363: Deprecating legacy function

src/types/account-manager.ts
```diff
@@ -285,6 +427,7 @@ export class AccountManager {
    * @returns The account
    */
   public async localNetDispenser() {
-    return this.signerAccount(await getLocalNetDispenserAccount(this._clientManager.algod, this._clientManager.kmd))
+    const dispenser = await this._kmdAccountManager.getLocalNetDispenserAccount()
```
robdmoore, line 430: Deprecating legacy function

src/types/account.ts
```diff
@@ -144,3 +131,18 @@ export type AccountAssetInformation = {
   /** The round as at which the holding was correct. */
   round: bigint
 }
+
+/**
+ * @deprecated The methods that use this can be achieved using `AccountManager` instead.
```
robdmoore, line 136: Now deprecated

src/types/client-manager.ts
```diff
@@ -32,15 +50,31 @@ export class ClientManager {
...
     const _clients =
       'algod' in clientsOrConfig
         ? clientsOrConfig
         : {
-            algod: getAlgoClient(clientsOrConfig.algodConfig),
-            indexer: clientsOrConfig.indexerConfig ? getAlgoIndexerClient(clientsOrConfig.indexerConfig) : undefined,
-            kmd: clientsOrConfig.kmdConfig ? getAlgoKmdClient(clientsOrConfig.kmdConfig) : undefined,
+            algod: ClientManager.getAlgodClient(clientsOrConfig.algodConfig),
+            indexer: clientsOrConfig.indexerConfig ? ClientManager.getIndexerClient(clientsOrConfig.indexerConfig) : undefined,
+            kmd: clientsOrConfig.kmdConfig ? ClientManager.getKmdClient(clientsOrConfig.kmdConfig) : undefined,
```
robdmoore, line 77: Deprecating legacy functions

src/types/client-manager.ts
```diff
@@ -64,14 +98,63 @@ export class ClientManager {
...
+  /**
+   * Get details about the current network.
+   * @example Getting genesis ID
+   * ```typescript
+   * const network = await networkClient.network()
+   * const genesisId = network.genesisId
+   * ```
+   * @returns The current network details
+   */
+  public async network(): Promise<NetworkDetails> {
```
robdmoore, line 111: This plus the following deprecates a series of legacy methods and has better performance since it caches the network params

src/types/client-manager.ts
```diff
@@ -130,6 +213,237 @@ export class ClientManager {
   public getTypedAppClientById<TClient>(typedClient: TypedAppClient<TClient>, details: TypedAppClientByIdDetails) {
     return new typedClient({ ...details, resolveBy: 'id' }, this._algod)
   }
+
```
robdmoore, line 216: Each of these static methods deprecate various legacy methods

src/types/client-manager.ts
```diff
@@ -130,6 +213,237 @@ export class ClientManager {
...
+    if (!process || !process.env) {
+      throw new Error('Attempt to get default client configuration from a non Node.js context; supply the config instead')
+    }
+    const algodConfig = !process.env.ALGOD_SERVER
+      ? ClientManager.getDefaultLocalNetConfig('algod')
+      : ClientManager.getAlgodConfigFromEnvironment()
+
+    const indexerConfig = !process.env.INDEXER_SERVER
+      ? ClientManager.getDefaultLocalNetConfig('indexer')
+      : !process.env.ALGOD_SERVER
```
robdmoore, line 245: This is technically a breaking change, but I figured it would be super weird if you used environment variables for algod, but were happy with default localnet for indexer / kmd so probably wouldn't break anyone in practice.

Given the concept of indexer and kmd being optional is more baked in now it felt better to make them undefined in the case you only define algod environment variables so you can support that explicitly.

src/types/client-manager.ts
```diff
@@ -130,6 +213,237 @@ export class ClientManager {
...
+   *
+   * @param network Which network to connect to - TestNet or MainNet
+   * @param config Which algod config to return - Algod or Indexer
+   */
+  public static getAlgoNodeConfig(network: 'testnet' | 'mainnet', config: 'algod' | 'indexer'): AlgoClientConfig {
+    return {
+      server: `https://${network}-${config === 'algod' ? 'api' : 'idx'}.algonode.cloud/`,
+      port: 443,
+    }
+  }
```
robdmoore, line 311: I saw a tweet the other day from Algorand Foundation that indicated there were 4 node services, we should create similar methods for any others that have a free tier?

robdmoore: Found it! https://twitter.com/algodevs/status/1796342119109407172

src/types/testing.ts
```diff
@@ -42,17 +45,17 @@ export interface GetTestAccountParams {
   initialFunds: AlgoAmount
   /** Whether to suppress the log (which includes a mnemonic) or not (default: do not suppress the log) */
   suppressLog?: boolean
-  /** Optional override for how to get an account; this allows you to retrieve accounts from a known or cached list of accounts. */
-  accountGetter?: (algod: Algodv2, kmd?: Kmd) => Promise<Account>
+  /** Optional override for how to get a test account; this allows you to retrieve accounts from a known or cached list of accounts. */
+  accountGetter?: (algorand: AlgorandClient) => Promise<Account>
```
robdmoore, line 49: Technically this is a breaking change, but it's a very recent feature that wasn't advertised and I'm pretty confident it's unlikely people will be using it.

src/types/account-manager.ts
neilcampbell: Fine by me.

src/types/kmd-account-manager.ts
```diff
@@ -0,0 +1,171 @@
...
+  public async getWalletAccount(
+    walletName: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    predicate?: (account: Record<string, any>) => boolean,
+    sender?: string,
+  ): Promise<(TransactionSignerAccount & { account: SigningAccount }) | undefined> {
+    const wallets = await this._clientManager.kmd.listWallets()
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const wallet = wallets.wallets.filter((w: any) => w.name === walletName)
```
neilcampbell, line 46:
```suggestion
    const walletsResponse = await this._clientManager.kmd.listWallets()

    // eslint-disable-next-line @typescript-eslint/no-explicit-any
    const wallet = walletsResponse.wallets.filter((w: any) => w.name === walletName)
```

src/types/kmd-account-manager.ts
```diff
@@ -0,0 +1,171 @@
...
+          break
+        }
+      }
+    }
+
+    if (i >= keyIds.length) {
+      return undefined
+    }
+
+    const accountKey = (await this._clientManager.kmd.exportKey(walletHandle, '', keyIds[i])).private_key
```
neilcampbell, line 71:
```suggestion
    const addresses = (await this._clientManager.kmd.listKeys(walletHandle)).addresses

    let i = 0
    if (predicate) {
      for (i = 0; i < addresses.length; i++) {
        const address = addresses[i]
        const account = await this._clientManager.algod.accountInformation(address).do()
        if (predicate(account)) {
          break
        }
      }
    }

    if (i >= addresses.length) {
      return undefined
    }

    const accountKey = (await this._clientManager.kmd.exportKey(walletHandle, '', addresses[i])).private_key
```

src/types/account-manager.ts
```diff
@@ -257,22 +388,33 @@ export class AccountManager {
...
   }
 
   /**
-   * Returns an account (with private key loaded) that can act as a dispenser.
+   * Returns an account (with private key loaded) that can act as a dispenser from
+   * environment variables, or against default LocalNet if no environment variables present.
+   *
+   * Note: requires a Node.js environment to execute.
+   *
+   * It present, it will load the account mnemonic stored in process.env.DISPENSER_MNEMONIC and optionally
```
neilcampbell, line 400:
```suggestion
   * If present, it will load the account mnemonic stored in process.env.DISPENSER_MNEMONIC and optionally
```

src/types/account-manager.ts
```diff
@@ -81,12 +134,17 @@ export class AccountManager {
...
    *
-   * If no signer has been registered for that address then the default signer is used if registered.
+   * If no signer has been registered for that address then the default signer is used if registered and
+   * if not then an error is thrown.
    *
    * @param sender The sender address
-   * @returns The `TransactionSigner` or throws an error if not found
+   * @example
+   * ```typescript
+   * const signer = accountManager.getSigner("{SENDERADDRESS}")
```
neilcampbell, line 145:
```suggestion
   * const signer = accountManager.getSigner("SENDERADDRESS")
```

src/network-client.ts
```diff
@@ -1,106 +1,65 @@
...
-
-  // Because we override the default HTTP Client construction (to get retries) we need to put a string token into the standard header ourselves
-  return typeof token === 'string' ? { [defaultHeader ?? 'X-Algo-API-Token']: token } : token ?? {}
-}
-
-/** Returns an algod SDK client that automatically retries on idempotent calls
+/**
+ * @deprecated Use `ClientManager.getAlgodClient(config)` or `ClientManager.getAlgodClientFromEnvironment()` instead.
+ *
+ * Returns an algod SDK client that automatically retries on idempotent calls
```
neilcampbell, line 62:
```suggestion
 * Returns an algod SDK client that automatically retries transient failures on idempotent calls
```

src/network-client.ts
```diff
@@ -128,12 +87,13 @@ function getAlgoTokenHeader(server: string, token?: string | TokenHeader, defaul
...
-  const httpClientWithRetry = new AlgoHttpClientWithRetry(getAlgoTokenHeader(server, token), server, port)
-  return new algosdk.Algodv2(httpClientWithRetry, server)
+  return config ? ClientManager.getAlgodClient(config) : ClientManager.getAlgodClientFromEnvironment()
 }
 
-/** Returns an indexer SDK client that automatically retries on idempotent calls
+/**
+ * @deprecated Use `ClientManager.getIndexerClient(config, overrideIntDecoding)` or `ClientManager.getIndexerClientFromEnvironment(overrideIntDecoding)` instead.
+ *
+ * Returns an indexer SDK client that automatically retries on idempotent calls
```
neilcampbell, line 96:
```suggestion
 * Returns an indexer SDK client that automatically retries transient failures on idempotent calls
```

src/network-client.ts
```diff
@@ -165,16 +125,15 @@ export function getAlgoClient(config?: AlgoClientConfig): Algodv2 {
...
+  return config
+    ? ClientManager.getIndexerClient(config, overrideIntDecoding)
+    : ClientManager.getIndexerClientFromEnvironment(overrideIntDecoding)
 }
 
 /**
- * Returns a KMD SDK client that automatically retries on idempotent calls
+ * @deprecated Use `ClientManager.getKmdClient(config)` or `ClientManager.getKmdClientFromEnvironment()` instead.
+ *
+ * Returns a KMD SDK client that automatically retries on idempotent calls.
```
neilcampbell, line 136:
```suggestion
 * Returns a KMD SDK client that automatically retries transient failures on idempotent calls.
```

src/index.ts
robdmoore: What about:

`import { AlgorandClient, Config } from 'algokit-utils/core'`

robdmoore: or we just avoid this and people deal with the noise and do:

`import { AlgorandClient, Config } from '@algorandfoundation/algokit-utils'`

joe-p:
LGTM. We should get eyes on it from devrel before merging though mainly to keep them in the loop and make sure there's no objections. 

@SilentRhetoric no need for line-by-line review, but general summary is

* Moving AlgokitComposer/AlgorandClient out of beta
* Marking older functions now covered by these classes as deprecated
* Updating documentation to document new ways of doing things and guide developers from old to new
* Implementing AlgorandClient deeper within the library in some places to reduce duplicate code

SilentRhetoric:
Thank you for your work on this! The consolidation of clients should make the library easier to use, investigate with Intellisense, teach, and maintain.

robdmoore:
Agreed! This is the first of a number of pull requests to gracefully bring the new way of working with the library into being so expect a few more pull requests.

neilcampbell (approved):

PR #291, opened 2024-06-17
feat: Deprecate transfer and asset modules

robdmoore:
Follows on from #287.

Breaking changes:
* Renamed `AlgokitComposer` to `AlgoKitComposer` to match AlgoKit naming conventions
* Collapsed the second object in algorand.send.type(params, *executeOptions*) to combine into params and make it easier to use based on devrel feedback
* Order of `algorand.account.rekeyed()` parameters to be `rekeyed(sender, signer)` since it conceptually makes more sense (the sender is rekeyed so should come first)
* All microAlgo return values from `algorand.account.getInformation()` now return an `AlgoAmount`, renamed `amount` to `balance` and `round` to `validAsAtRound` (which is now a `bigint` for broader consistency)
* Renamed `algorand.account.getAssetInformation` to `algorand.asset.getAccountInformation`

Added:
* Return value of `algorand.send.assetCreate` now includes `{ assetId: bigint }`
* Up to date documentation for rekeyAccount, AlgorandClient, asset, dispenser client, indexer, testing, transfer
* Up to date documentation for how to refer to an Algo amount (Algo (not plural) in general and ALGO (not plural) when referring to a specific amount)
* `AssetManager` class and `algorand.asset.getById()`, `algorand.asset.bulkOptIn`, and `algorand.asset.bulkOptOut`
* `indexer` export off of `@algorandfoundation/algokit-utils` as the future home of all indexer methods
* Added `algorand.client.getTestNetDispenserFromEnvironment`
* Added `algorand.account.assetBulkOptIn`
* Added `algorand.account.assetBulkOptOut`
* Added `algorand.account.ensureFunded`
* Added `algorand.account.ensureFundedFromEnvironment`
* Added `algorand.account.ensureFundedFromTestNetDispenserApi`
* Added `algorand.account.rekeyAccount`
* Added `algorand.send/transaction.assetOptOut`
* Added `buildTransactions` method to `AlgoKitComposer` so you can build transactions without needing to register a signer
* Added `.algo` and `.microAlgo` methods/properties in place of the plural version for `AlgoAmount` to reflect the current guidance on how to represent Algo amounts, kept previous versions of those methods for now to avoid the breaking change

Deprecated the following in favour of AlgorandClient functionality:
* `algokit.createAsset`
* `algokit.assetOptIn`
* `algokit.assetOptOut`
* `algokit.assetBulkOptIn`
* `algokit.assetBulkOptOut`
* `algokit.ensureFunded`
* `algokit.transferAsset`
* `algokit.rekeyAccount`
* `algokit.transferAlgos`


docs/capabilities/algorand-client.md
```diff
@@ -36,3 +36,31 @@ const algodClient = algorand.client.algod
...
+  - `algorand.setSignerFromAccount(account)` -
+  - `algorand.setSigner(sender, signer)`
+- [`AssetManager`](./asset.md) via `algorand.asset`
+- [`ClientManager`](./client.md) via `algorand.client`
+
+## Creating and issuing transactions
+
+`AlgorandClient` exposes a series of methods that allow you to create, execute, and compose groups of transactions:
+
+### Creating single transactions
```
robdmoore, line 55: WIP

src/account/account.ts
```diff
@@ -144,7 +150,26 @@ export function getAccountAddressAsString(addressEncodedInB64: string): string {
  * @returns The account information
  */
 export async function getAccountInformation(sender: string | SendTransactionFrom, algod: Algodv2): Promise<AccountInformation> {
-  return new AccountManager(new ClientManager({ algod })).getInformation(getSenderAddress(sender))
+  const account = AccountInformationModel.from_obj_for_encoding(await algod.accountInformation(getSenderAddress(sender)).do())
```
robdmoore, line 153: Keeping backwards compatibility in the legacy method (made changes to this method in `AccountManager` to return better types (`AlgoAmount` rather than `number`).

src/index.ts
```diff
@@ -2,19 +2,20 @@
...
 export * from './dispenser-client'
+export * from './indexer-lookup'
 export * from './localnet'
 export * from './network-client'
+export * from './transfer'
 
 // Up to date exports
 export * from './amount'
 export * from './config'
+export * as indexer from './indexer-lookup'
```
robdmoore, line 20: I think the indexer methods are useful as-is. Per previous discussions we don't want to make them that easy to use given preference to discourage use of indexer. I figure exposing them through an `indexer` export off the `/` is a good way of exposing them without making for a noisy `/` import.

src/testing/_asset.ts
```diff
@@ -1,38 +1,23 @@
-import algosdk from 'algosdk'
-import Account = algosdk.Account
-import Algodv2 = algosdk.Algodv2
-import makeAssetCreateTxnWithSuggestedParamsFromObject = algosdk.makeAssetCreateTxnWithSuggestedParamsFromObject
+import { AlgorandClient } from '../types/algorand-client'
 
-export async function generateTestAsset(client: Algodv2, sender: Account, total?: number) {
+export async function generateTestAsset(algorand: AlgorandClient, sender: string, total?: number) {
```
robdmoore, line 3: Internal method so it's a non-breaking change for the library.

src/types/algorand-client.asset.spec.ts
```diff
@@ -0,0 +1,198 @@
+import { describe, test } from '@jest/globals'
```
robdmoore, line 1: Rename of asset.spec.ts with changes to use algorandClient

src/types/algorand-client.transfer.spec.ts
```diff
@@ -0,0 +1,378 @@
+import { describe, test } from '@jest/globals'
```
robdmoore, line 1: Rename of transfer.spec.ts with changes to use algorandClient

src/types/algorand-client.ts
```diff
@@ -118,8 +135,13 @@ export class AlgorandClient {
     return this._accountManager
   }
 
+  /** Methods for interacting with assets. */
+  public get asset() {
+    return this._assetManager
```
robdmoore, line 140: Added new manager (assetManager). For now it's simple (single method), but over time it may grow in complexity.

src/types/algorand-client.ts
```diff
@@ -134,19 +156,19 @@ export class AlgorandClient {
       preLog?: (params: T, transaction: Transaction) => string
       postLog?: (params: T, result: SendSingleTransactionResult) => string
     },
-  ): (params: T, config?: ExecuteParams) => Promise<SendSingleTransactionResult> {
-    return async (params, config) => {
+  ): (params: T & ExecuteParams) => Promise<SendSingleTransactionResult> {
```
robdmoore, line 159: Combining these together per devrel feedback

joe-p:
Haven't done a proper review yet, but not sure how I feel about `algorand.send/transaction.rekey`. On one hand it's nice to have such an important operation be so explicit. On the other hand, I'm worried it might give developers the incorrect impression that a rekey is a special transaction type.

robdmoore:
No different to opt-in/opt-out imo

robdmoore:
Also allows for a really explicit approach to such a dangerous but useful operation with a minimal set of parameters with intellisense  we could illustrate in the method description that rekey can also be added to any other transaction by setting the optional rekeyTo value?

joe-p:
> No different to opt-in/opt-out imo

To me the main difference is that there are no negative consequences of a developer thinking optin/out is a seperate tx type. With rekeys, however, if a developer thinks that rekey is a special transaction type and reads code somewhere with a payment transaction or appl, they might be oblivious to the fact that it's rekeying an account.

Essentially I'm nervous it might "train" developers to look for `.rekey` calls rather than `rekeyTo` fields

robdmoore:
Another option is to add it to account manager as a method.

joe-p:
> Another option is to add it to account manager as a method.

Yeah I like that. Makes it more clear that it's abstraction and not a transaction type

robdmoore:
Cool, done.

Funnily enough it was how I originally wrote it, but then I decided to make it a composer thing so you could get the transaction separate to sending it. I think that's a minor benefit though.

joe-p:
small nit: I would change `rekeyAccount` to `rekeyAddress`. In general as we start to get HD wallet support we should try to avoid using "account" to refer to a single address.

docs/capabilities/account.md
```diff
@@ -62,31 +62,53 @@ In order to get/register accounts for signing operations you can use the followi
...
+  //  already specified, but here for completeness
+  maxFee: (3000).microAlgos(),
+  maxRoundsToWaitForConfirmation: 5,
+  suppressLog: true,
+})
+
+// Using a rekeyed account
 
+// Note: if a signing account is passed into `algorand.account.rekeyAccount` then you don't need to call `rekeyedAccount` to register the new signer
 const rekeyedAccount = algokit.rekeyedAccount(newAccount, account.addr)
```
joe-p, line 112:
```suggestion
const rekeyedAccount = Algorand.account.rekeyed(newAccount, account.addr)
```

docs/capabilities/algorand-client.md
```diff
@@ -2,7 +2,7 @@
 
 `AlgorandClient` is a client class that brokers easy access to Algorand functionality. It's the [default entrypoint](../README.md#usage) into AlgoKit Utils functionality.
 
-The main entrypoint to the bulk of the functionality in AlgoKit Utils is the `AlgorandClient` class, most of the time you can get started by typing `algokit.AlgorandClient.` and choosing one of the static initialisation methods to create an [Algorand client](./capabilities/algorand-client.md), e.g.:
+The main entrypoint to the bulk of the functionality in AlgoKit Utils is the `AlgorandClient` class, most of the time you can get started by typing `algokit.AlgorandClient.` and choosing one of the static initialisation methods to create an [Algorand client](../code/classes/types_algorand_client.AlgorandClient.md), e.g.:
```
joe-p, line 5: Aren't we now recommending to import AlgorandClient directly?

src/indexer-lookup.spec.ts
```diff
@@ -65,9 +58,9 @@ describe('indexer-lookup', () => {
...
 
   test('Application create transactions are found by creator with pagination', async () => {
-    const { algod, indexer, testAccount, generateAccount, waitForIndexer } = localnet.context
+    const { algorand, testAccount, generateAccount, waitForIndexer } = localnet.context
     const secondAccount = await generateAccount({
-      initialFunds: algokit.algos(1),
+      initialFunds: (1).algos(),
       suppressLog: true,
     })
     const createParams = await getTestingAppCreateParams(testAccount, {
```
joe-p, line 66: This line can be deleted

src/types/algorand-client.ts
joe-p: Should the bulk functions go in the AssetManager?

joe-p (commented):
@robdmoore can you remind me why we decided to support a union of `string | TransactionSignerAccount` for address fields? When would the `TransactionSignerAccount` be useful? I think it might be a bit confusing especially when I was looking at `account.rekeyed`. I suppose it's not clear when I'd want to use this vs setting the signer for address. 

Also, I see a lot of changes to the formatting of how we saw "Algos". Personally I think non-plural ALGO is the most common. Ie. "I have 10 ALGO" vs "I have 10 Algos"

robdmoore:
> why we decided to support a union of string | TransactionSignerAccount for address fields?

It was so that you could pass in the result of something like `const alice = algorand.account.random()` in i.e. you can pass in `alice` and not have to do `alice.addr` i.e. `algorand.send.payment({from: alice, to: bob})` reads much nicer than `algorand.send.payment({from: alice.addr, to: bob.addr})`

We could remove that and just stick with string of course if we felt it was too confusing...

robdmoore:
> Also, I see a lot of changes to the formatting of how we saw "Algos". Personally I think non-plural ALGO is the most common. Ie. "I have 10 ALGO" vs "I have 10 Algos"

It's a tough one because I see it written in different ways everywhere, it's wildly inconsistent. I'm not opposed to just using ALGO though, more than happy with that?

robdmoore:
> Should the bulk functions go in the AssetManager?

I thought about that. Maybe that makes sense? I put in account because it was being done in the context of an account, but `AccountManager` is a really big class now and you could argue either way, so maybe `AssetManager` is better?

docs/capabilities/algorand-client.md
robdmoore: Yep this needs to be fixed

joe-p:
> It was so that you could pass in the result of something like const alice = algorand.account.random()

This begs the question... why not just have `.random` return a string? Returning an object with the signer makes sense when we are concerned about legacy interop, but now that we're deprecating legacy utils functions maybe it just makes more sense to return a string? I suppose the question really is where do we expect developers to use the signer function manually rather than using it implicitly through the AlgorandClient


> It's a tough one because I see it written in different ways everywhere, it's wildly inconsistent. I'm not opposed to just using ALGO though, more than happy with that?

Yeah I don't have a strong opinion on it but consistency is nice at least within the libs we control. Perhaps we can just do a quick poll in slack to see what everyone prefers and just stick with that everywhere. 

> I thought about that. Maybe that makes sense? I put in account because it was being done in the context of an account, but AccountManager is a really big class now and you could argue either way, so maybe AssetManager is better?

Yeah I can see the argument either way. I personally leans towards AssetManager but have no objections for AccountManager. Although since AccountManager is already a big class and these are relatively uncommon functions to use, it might make more sense to put under AssetManager just for that reason.

robdmoore:
I like the simplicity of returning a string, but it feels very wrong from an object oriented / code semantics perspective. ‘String’ is too generic a type for something that important and it would make it much harder to detect when you’re passing the wrong thing in the wrong place maybe?

agree on the other too. Did you want to raise the poll for algos?

joe-p:
> I like the simplicity of returning a string, but it feels very wrong from an object oriented / code semantics perspective. ‘String’ is too generic a type for something that important and it would make it much harder to detect when you’re passing the wrong thing in the wrong place maybe?

I agree but also think that's just the nature of nominal type systems. I'm not convinced it's worth creating an object just for this purpose. 

> agree on the other two. Did you want to raise the poll for algos?

Just posted in slack 👍

src/types/algorand-client.ts
```diff
@@ -242,21 +600,276 @@ export class AlgorandClient {
...
+     *
+     * @param params The parameters for the asset config transaction
+     *
+     * @example Basic example
+     * ```typescript
+     * await algorand.transaction.assetConfig({sender: "MANAGERADDRESS", assetId: 123456n, manager: "MANAGERADDRESS" })
+     * ```
+     * @example Advanced example
+     * ```typescript
+     * await algorand.transaction.assetCreate({
```
neilcampbell, line 695:
```suggestion
     * await algorand.transaction.assetConfig({
```

docs/capabilities/account.md
```diff
@@ -47,10 +47,10 @@ In order to get/register accounts for signing operations you can use the followi
 
 - [`algorand.account.fromEnvironment(name, fundWith)`](../code/classes/types_account_manager.AccountManager.md#fromenvironment) - Registers and returns an account with private key loaded by convention based on the given name identifier - either by idempotently creating the account in KMD or from environment variable via `process.env['{NAME}_MNEMONIC']` and (optionally) `process.env['{NAME}_SENDER']` (if account is rekeyed)
   - This allows you to have powerful code that will automatically create and fund an account by name locally and when deployed against TestNet/MainNet will automatically resolve from environment variables, without having to have different code
-  - Note: `fundWith` allows you to control how many ALGOs are seeded into an account created in KMD
+  - Note: `fundWith` allows you to control how many Algos are seeded into an account created in KMD
```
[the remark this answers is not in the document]
robdmoore: Done

robdmoore:
Regardless of whether we decide to do string or not, it makes sense to decouple that from this PR since it's a much wider / bigger / unrelated change.

I also didn't change rekeyAccount to rekeyAddress because there are many other things that tie account = single address, that's a broader change that will somehow need to be incorporated into the Algorand ecosystem with thought to avoid confusion.

Otherwise incorporated all the other feedback.

src/transaction/transaction.ts
```diff
@@ -807,7 +807,7 @@ export const waitForConfirmation = async function (
 }
 
 /**
- * Limit the acceptable fee to a defined amount of µALGOs.
+ * Limit the acceptable fee to a defined amount of µAlgo.
```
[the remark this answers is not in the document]
robdmoore: This was deliberate. The guidance was that you use ALGO when there is a specific amount, but otherwise it's referred to as Algo

src/types/algorand-client-sender.ts
```diff
@@ -0,0 +1,459 @@
...
+export type SendSingleTransactionResult = SendAtomicTransactionComposerResults & ConfirmedTransactionResult
+
+/** Orchestrates sending transactions for `AlgorandClient`. */
+export class AlgorandClientSender {
+  private _newGroup: () => AlgoKitComposer
+  private _assetManager: AssetManager
+
+  /**
+   * Creates a new `AlgorandClientSender`
+   * @param newGroup A lambda that Start a new `AlgokitComposer` transaction group
```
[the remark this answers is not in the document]
robdmoore: No, see other comment

src/types/algorand-client-transaction-creator.ts
```diff
@@ -0,0 +1,304 @@
+import algosdk from 'algosdk'
+import AlgoKitComposer, { MethodCallParams } from './composer'
+
+import Transaction = algosdk.Transaction
+
+/** Orchestrates sending transactions for `AlgorandClient`. */
```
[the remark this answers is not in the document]
robdmoore: 👍

src/types/algorand-client-transaction-creator.ts
```diff
@@ -0,0 +1,304 @@
...
+export class AlgorandClientTransactionCreator {
+  private _newGroup: () => AlgoKitComposer
+
+  /**
+   * Creates a new `AlgorandClientSender`
+   * @param newGroup A lambda that Start a new `AlgokitComposer` transaction group
```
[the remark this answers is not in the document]
robdmoore: 👍

src/types/algorand-client-transaction-creator.ts
```diff
@@ -0,0 +1,304 @@
...
+import AlgoKitComposer, { MethodCallParams } from './composer'
+
+import Transaction = algosdk.Transaction
+
+/** Orchestrates sending transactions for `AlgorandClient`. */
+export class AlgorandClientTransactionCreator {
+  private _newGroup: () => AlgoKitComposer
+
+  /**
+   * Creates a new `AlgorandClientSender`
```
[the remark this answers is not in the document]
robdmoore: 👍

src/types/algorand-client-transaction-creator.ts
```diff
@@ -0,0 +1,304 @@
...
+   * @param newGroup A lambda that Start a new `AlgokitComposer` transaction group
+   * @param assetManager An `AssetManager` instance
```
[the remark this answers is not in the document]
robdmoore: 👍

neilcampbell (approved):

PR #292, opened 2024-06-21
feat: first/last valid on SendTransactionParams

joe-p:
## Proposed Changes

Simply adds first and last valid to SendTransactionParams


src/transaction/transaction.ts
```diff
@@ -194,6 +194,9 @@ export const sendTransaction = async function (
   const { transaction, from, sendParams } = send
   const { skipSending, skipWaiting, fee, maxFee, suppressLog, maxRoundsToWaitForConfirmation, atc } = sendParams ?? {}
 
+  if (sendParams?.firstValid) transaction.firstRound = sendParams.firstValid
```
robdmoore, line 197: Should we also add it to sendAtomicTransactionComposer, from memory it's a different code path, but a bit hazy on it.

Also, we should add this to the legacyTransactionBridge in #287

joe-p: Well we can't really add it to `sendAtomicTransactionComposer` because it's a transaction-specific field. Setting it on `sendAtomicTransactionComposer` would basically force us to update ALL the transactions with the specified lv/fv which we might not want.

robdmoore: Yeah, that makes sense, I think I'm just getting confused about the code paths

robdmoore (approved):

src/types/transaction.ts
```diff
@@ -43,6 +43,10 @@ export interface SendTransactionParams {
   maxRoundsToWaitForConfirmation?: number
   /** **WARNING**: Not recommended for production use due to https://github.com/algorand/go-algorand/issues/5914. Whether to use simulate to automatically populate app call resources in the txn objects. Defaults to true when there are app calls in the group.  */
   populateAppCallResources?: boolean
+  /** The first round this transaction will be valid */
+  firstValid?: number
+  /** The last round this transaction will be valid */
+  lastValid?: number
```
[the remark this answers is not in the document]
joe-p: Yeah good point. Let me know what you think of [623edb0](https://github.com/algorandfoundation/algokit-utils-ts/pull/292/commits/623edb0e3a0396ae9b89c37b5331e07b88f72f7e)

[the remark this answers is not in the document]
robdmoore: Looks good, noting most methods also allow you to pass in suggested params as an override to avoid the algod call, but maybe that detail doesn't matter

src/transaction/transaction.ts
```diff
@@ -217,6 +217,11 @@ export const sendTransaction = async function (
   const { transaction, from, sendParams } = send
   const { skipSending, skipWaiting, fee, maxFee, suppressLog, maxRoundsToWaitForConfirmation, atc } = sendParams ?? {}
 
+  // @ts-expect-error firstValid is marked as read only
+  if (sendParams?.firstValid) transaction.firstValid = sendParams.firstValid
+  // @ts-expect-error lastValid is marked as read only
+  if (sendParams?.lastValid) transaction.lastValid = sendParams.lastValid
+
```
joe-p, line 224: @lempira After merging main (presumably because of SDK version bump) I had to use `@ts-expect-error` since the fields are marked as read-only but we need to change them

lempira: Would we still get this if we deep clone the `sendParams`?

lempira (commented):
Just had the question in the comment I left

PR #297, opened 2024-07-12
fix: local and global state delta types

negar-abbasi:
## Proposed Changes

- update the types for global state delta and local state **delta**
-
-


neilcampbell (approved):

neilcampbell:
This would typically be a breaking change, however we're correcting an incorrect type so I think it's ok to release as non breaking. Presumably anyone using this would be retyping.

PR #300, opened 2024-07-19
fix: Ensuring that transactions issued by AlgoKit Composer to LocalNet don't have too short a validity window

robdmoore:


robdmoore:
Fixes #301

PR #303, opened 2024-07-19
fix: match the previous getTestAccount behaviour when no kmd client is supplied

neilcampbell:
Realised I hadn't confirmed the new behaviour when kmd wasn't supplied.
Turns out it throws, which wasn't as per the documented behaviour.

This PR updates the obsoleted `getTestAccount` to have the same behaviour as before.

### DIFF
--- a/src/testing/account.ts
+++ b/src/testing/account.ts
@@ -1,5 +1,6 @@
 import algosdk from 'algosdk'
 import { AlgorandClient, Config } from '../'
+import { ClientManager } from '../types/client-manager'
 import { GetTestAccountParams } from '../types/testing'
 import Account = algosdk.Account
 import Algodv2 = algosdk.Algodv2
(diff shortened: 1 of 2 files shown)


src/testing/account.ts
```diff
@@ -36,7 +37,11 @@ export async function getTestAccount(
   const algorand =
     algodOrAlgorandClient instanceof AlgorandClient
       ? algodOrAlgorandClient
-      : AlgorandClient.fromClients({ algod: algodOrAlgorandClient, kmd })
+      : AlgorandClient.fromClients({
+          algod: algodOrAlgorandClient,
+          kmd:
```
robdmoore, line 42: `ClientManager.getConfigFromEnvironmentOrLocalNet` is probably better to avoid the duplication of KMD logic.

neilcampbell: I was thinking of using this, however previous code didn't have a LocalNet fallback, so it's behaviourally different. Given it should only be used for testing, it should be safe.

neilcampbell: Updated.

robdmoore (approved):

PR #304, opened 2024-08-09
feat: return/accept string as address in AccountManager methods

joe-p:
## Proposed Changes

- All AccountManager methods now accept/return a string. The idea being that most users don't have to think about signers/account objects since all the places they use account accept a string as an argument and the AccountManager will figure out the signer automatically

## Discussion Points

* It should be noted that this now creates a discrepancy between the test fixture and AccountManager methods, since the test fixture returns account objects. I'd propose that we essentially create new test fixture objects and deprecate the current ones, but open to other ideas.


neilcampbell:
As discussed internally utils-ts@8 is moving to an approach where the algosdk@3 Address type is returned, so closing in favour of that approach.

PR #312, opened 2024-09-05
fix: bigint template variables

joe-p:
## Proposed Changes

- Fixes bigint values for template variables when pushbytes is used

### DIFF
--- a/src/types/app-manager.ts
+++ b/src/types/app-manager.ts
@@ -434,7 +434,7 @@ export class AppManager {
 
         // If this is a number, first replace any byte representations of the number
         // These may appear in the TEAL in order to circumvent int compression and preserve PC values
-        if (typeof value === 'number' || typeof value === 'boolean') {
+        if (typeof value === 'number' || typeof value === 'bigint') {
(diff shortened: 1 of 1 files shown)


src/types/app-manager.ts
```diff
@@ -434,7 +434,7 @@ export class AppManager {
 
         // If this is a number, first replace any byte representations of the number
         // These may appear in the TEAL in order to circumvent int compression and preserve PC values
-        if (typeof value === 'number' || typeof value === 'boolean') {
```
robdmoore, line 437: Boolean still makes sense (converts to `0`/`1`)

robdmoore: Thoughts @joe-p?

joe-p: I suppose it depends on if we expect booleans to be ABI encoded or not. How does Puya currently handle boolean template variables?

robdmoore: If we want ABI encoded the current implementation would work if you passed through a uint8array rather than a boolean. TBF maybe we should only allow bigint and uint8array at this low level anyway, and then it's the higher level app client construct that does the fancy stuff, that matches the way I've implemented the rest of app client stuff vs composer.

robdmoore: I just realised we don't even allow boolean anyway.

robdmoore (approved):

PR #317, opened 2024-09-27
V7 tweaks

robdmoore:
Tweaks based on feedback from various people including @neilcampbell and @SilentRhetoric to prepare for v7 finalisation


docs/capabilities/app-client.md
```diff
@@ -108,9 +108,9 @@ The create method is a wrapper over the `appCreate` (bare calls) and `appCreateM
 
 ```typescript
 // Use no-argument bare-call
-const { result, app } = factory.create()
+const { result, appClient } = factory.send.bare.create()
```
robdmoore, line 111: @SilentRhetoric even though it adds a little bit of verbosity we made the decision that every time a transaction is sent to the network you will see the word `send` (apart from the `deploy` call, but that's a special case since it may or may not send transactions and is a higher-order function).

As part of this we renamed `execute` to `send` (with a backwards compatible `@deprecate`) in `AlgoKitComposer` and renamed `ExecuteParams` to `SendParams`.

docs/capabilities/app-client.md
```diff
@@ -108,9 +108,9 @@ The create method is a wrapper over the `appCreate` (bare calls) and `appCreateM
...
 // Specify parameters for bare-call and override other parameters
-const { result, app } = factory.create({
+const { result, appClient } = factory.send.bare.create({
```
robdmoore, line 113: We also settled on `appClient` over `app` to meet your principle of clear, explicit naming @SilentRhetoric

docs/capabilities/app-client.md
```diff
@@ -123,13 +123,13 @@ const { result, app } = factory.create({
...
 // Specify parameters for ABI method call
-const { result, app } = factory.create({
+const { result, appClient } = factory.send.create({
   method: 'create_application',
   args: [1, 'something'],
 })
 ```
 
-If you want to construct a custom create call, use the underlying [`algorand.send.appCreate` / `algorand.transactions.appCreate` / `algorand.send.appCreateMethodCall` / `algorand.transactions.appCreateMethodCall` methods](./app.md#creation) then you can get params objects:
+If you want to construct a custom create call, use the underlying [`algorand.send.appCreate` / `algorand.createTransaction.appCreate` / `algorand.send.appCreateMethodCall` / `algorand.createTransaction.appCreateMethodCall` methods](./app.md#creation) then you can get params objects:
```
robdmoore, line 132: `transactions` -> `createTransaction` to get the verb plus the consistency of singular rather than plural.

We **didn't** rename `params` because the _correct_ name would be `getTransactionParams` and it's simply too verbose and the places you are most likely to use params are inline when nesting ABI method arguments where you want less verbosity. As such we updated the jsdoc comment to make it more obvious how it's used and feel it's one that people will follow examples to understand how to use (and honestly, it's a slightly more advanced use case anyway so I think it's OK if it's a bit less discoverable).

src/testing/fixtures/algorand-fixture.ts
```diff
@@ -92,7 +92,7 @@ export function algorandFixture(fixtureConfig?: AlgorandFixtureConfig, config?:
     const transactionLoggerAlgod = transactionLogger.capture(algod)
     algorand = AlgorandClient.fromClients({ algod: transactionLoggerAlgod, indexer, kmd })
     const acc = await getTestAccount({ initialFunds: fixtureConfig?.testAccountFunding ?? algos(10), suppressLog: true }, algorand)
-    algorand.setSignerFromAccount(acc).setSuggestedParamsTimeout(0)
+    algorand.setSignerFromAccount(acc).setSuggestedParamsCacheTimeout(0)
```
robdmoore, line 95: @SilentRhetoric while reviewing the method names we realised this method name was more descriptive as `CacheTimeout`

src/transaction/transaction.ts
```diff
@@ -571,14 +571,12 @@ export async function populateAppCallResources(atc: algosdk.AtomicTransactionCom
...
- * @param atcSend The parameters controlling the send, including:
- *  * `atc` The `AtomicTransactionComposer`
- *  * `executeParams` The parameters to control the send behaviour
+ * @param atcSend The parameters controlling the send, including `atc` The `AtomicTransactionComposer` and params to control send behaviour
  * @param algod An algod client
  * @returns An object with transaction IDs, transactions, group transaction ID (`groupTransactionId`) if more than 1 transaction sent, and (if `skipWaiting` is `false` or unset) confirmation (`confirmation`)
  */
 export const sendAtomicTransactionComposer = async function (atcSend: AtomicTransactionComposerToSend, algod: Algodv2) {
-  const { atc: givenAtc, sendParams, executeParams } = atcSend
+  const { atc: givenAtc, sendParams, ...executeParams } = atcSend
```
robdmoore, line 579: As part of renaming to `SendParams` from `ExecuteParams` we also collapsed the values into the params object for `sendAtomicTransactionComposer`, which meets our new single param object style. This is a bit of a behind the scenes method anyway that's unlikely to be used broadly outside of utils itself and adding `executeParams` is a new thing in v7 beta anyway so this isn't a breaking change from v6.

src/types/app-arc56.ts
```diff
@@ -2,22 +2,26 @@ import algosdk from 'algosdk'
...
-        type: algosdk.ABIArgumentType
-      }
-    >
-  >
-  override readonly returns: Expand<
-    Omit<Method['returns'], 'type'> & {
-      type: algosdk.ABIReturnType
-    }
-  >
+  override readonly args: Array<Arc56MethodArg>
```
robdmoore, line 23: In various places we removed inline use of `Expand<>` since we found it caused some confusion when consuming from npm. Instead we have dedicated exported types that use Expand.

src/types/app-factory.ts
```diff
@@ -170,44 +179,95 @@ export class AppFactory {
...
 
-  /**
-   * Creates an instance of the app and returns the result of the creation
-   * transaction and an app client to interact with that app instance.
-   *
-   * Performs deploy-time TEAL template placeholder substitutions (if specified).
-   * @param params The parameters to create the app
-   * @returns The app client and the result of the creation transaction
-   */
-  public async create(params?: AppFactoryCreateParams) {
```
robdmoore, line 186: Rather than a `create` method that takes bare or abi we split in 2 for consistency with app client. Makes the method less confusing to consume too.

neilcampbell (approved):

PR #330, opened 2024-10-18
fix: add app client support for nested method call

neilcampbell:
Ensure the correct transaction group is built when an ABI method call has an ABI method call argument.


src/types/composer.ts
```diff
@@ -832,15 +834,25 @@ export default class AlgoKitComposer {
...
     }
 
-    for (let i = 0; i < (params.args ?? []).length; i++) {
+    for (let i = (params.args ?? []).length - 1; i >= 0; i--) {
       const arg = params.args![i]
       if (arg === undefined) {
+        // An undefined transaction argument signals that the value will be supplied by a method call argument
+        if (algosdk.abiTypeIsTransaction(params.method.args[i].type) && transactionsForGroup.length > 0) {
+          // Move the last transaction from the group to the method call arguments to appease algosdk
+          const placeholderTransaction = transactionsForGroup.splice(-1, 1)[0]
```
robdmoore, line 851: `splice` mutates the array - desired?

src/types/composer.ts
```diff
@@ -1218,6 +1236,8 @@ export default class AlgoKitComposer {
     return { atc: this.atc, transactions: this.atc.buildGroup(), methodCalls: this.atc['methodCalls'] }
   }
 
+  // TODO: NC - Docs about the behaviour
```
robdmoore, line 1239: todo

robdmoore (approved):

src/types/composer.ts
neilcampbell: Yep, it is.

PR #332, opened 2024-10-21
chore: update algosdk2 to the latest version

neilcampbell:
algosdk@2.9.0 has support in their simulate models for `fixSigners`, which we want to use in lora. See https://github.com/algorand/go-algorand/pull/5942


robdmoore (approved):

neilcampbell:
Closing, as we can't do this change until nodes have updated to algod@3.26. This change will also require additionally setting `fixSigners` whenever we set `allowEmptySignatures`. Additionally we can remove the `signerWithFixedSgnr` workaround.

PR #333, opened 2024-10-22
feat: support latest ARC56 source mapping (including cblock offset)

joe-p:
## Proposed Changes

- Support latest ARC56 spec

Note that there is more work that could be done here to reduce API calls, notably cahcing the program and/or constant offset if it is known


joe-p:
Moved to draft until I add a test for this

HashMapsData2Value:
Are all the .md files automatically generated?

HashMapsData2Value:
Btw https://github.com/algorandfoundation/ARCs/pull/258/files is still not yet merged, why merge this in before then?

tests/example-contracts/arc56_templates/artifacts/Templates.arc56_draft.json
```diff
@@ -0,0 +1,714 @@
...
+    "create": [],
+    "call": []
+  },
+  "sourceInfo": {
+    "approval": {
+      "sourceInfo": [
+        {
+          "teal": 11,
+          "source": "tests/contracts/general.algo.ts:24",
+          "pc": [
```
HashMapsData2Value, line 108: Is tests/contracts/general.algo.ts automatically generated? The path in the repo seems to be tests/example-contracts

joe-p:
> I assume all the .md files automatically generated? Or should a review inspect each line?

Yeah they are generated by TypeDoc so no need for manual review

joe-p:
> Btw https://github.com/algorandfoundation/ARCs/pull/258/files is still not yet merged, why merge this in before then?

This will just be going into the beta release, which adds ARC56 suppport. Right now the main consumer of ARC56 is reti staking platform

tests/example-contracts/arc56_templates/artifacts/Templates.arc56_draft.json
joe-p: Ah yeah I copied over the artifact from TEALScript, I need to bump the TEALScript version and re-generate it here in this repo so this path is correct

robdmoore (approved):

PR #337, opened 2024-11-05
feta: support fix signers in simulate

neilcampbell:
Now that 3.26 is on mainnet, we can use `fixSigners` when running a simulate.
This PR also removes the previous workaround.

### DIFF
--- a/src/account/account.ts
+++ b/src/account/account.ts
@@ -166,10 +166,12 @@ export async function getAccountInformation(sender: string | SendTransactionFrom
     totalAssetsOptedIn: Number(account.totalAssetsOptedIn),
     totalCreatedApps: Number(account.totalCreatedApps),
     totalCreatedAssets: Number(account.totalCreatedAssets),
-    appsTotalExtraPages: account.appsTotalExtraPages ? Number(account.appsTotalExtraPages) : undefined,
-    rewardBase: account.rewardBase ? Number(account.rewardBase) : undefined,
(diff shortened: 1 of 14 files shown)


src/account/account.ts
```diff
@@ -166,10 +166,12 @@ export async function getAccountInformation(sender: string | SendTransactionFrom
     totalAssetsOptedIn: Number(account.totalAssetsOptedIn),
     totalCreatedApps: Number(account.totalCreatedApps),
     totalCreatedAssets: Number(account.totalCreatedAssets),
-    appsTotalExtraPages: account.appsTotalExtraPages ? Number(account.appsTotalExtraPages) : undefined,
-    rewardBase: account.rewardBase ? Number(account.rewardBase) : undefined,
-    totalBoxBytes: account.totalBoxBytes ? Number(account.totalBoxBytes) : undefined,
-    totalBoxes: account.totalBoxes ? Number(account.totalBoxes) : undefined,
+    appsTotalExtraPages: account.appsTotalExtraPages !== undefined ? Number(account.appsTotalExtraPages) : undefined,
```
robdmoore, line 169: This stuff also should be added to AccountManager.getIfnormation

robdmoore (approved):

src/account/account.ts
neilcampbell: Will add as `lastHeartbeatRound` and `lastProposedRound` and make `bigint` to match semantics of `validAsOfRound` on that type.

PR #343, opened 2024-11-25
Convert v7 to vitest and update dependencies to match what v8 will have

neilcampbell:
This is to make the parallel management of 2 version/branches easier in the future.


.vscode/extensions.json
```diff
@@ -1,3 +1,3 @@
 {
-  "recommendations": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode", "orta.vscode-jest"]
+  "recommendations": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode"]
```
PatrickDinh, line 2: Should we add `vitest.explorer` into the list?

PatrickDinh (approved):

.vscode/extensions.json
neilcampbell: Yeah good call.

PR #353, opened 2024-12-13
feat: logic error logging

joe-p:
## Proposed Changes

- Logs logic error stack to via `Config.logger.error`

This is useful in test environments because some frameworks (jest, vitest) cut off errors. This means the framework will only show the unparsed logic error with the PC and opcodes. This PR adds explicit logging with the stack from the LogicError to ensure the developer gets feedback on where the error occurs

### DIFF
--- a/src/transaction/transaction.spec.ts
+++ b/src/transaction/transaction.spec.ts
@@ -441,9 +441,13 @@ describe('Resource Packer: meta', () => {
   })
 
   test('error during simulate', async () => {
-    await expect(externalClient.send.call({ method: 'error' })).rejects.toThrow(
-      'Error during resource population simulation in transaction 0',
-    )
+    try {
+      await externalClient.send.call({ method: 'error' })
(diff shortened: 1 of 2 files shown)


src/transaction/transaction.spec.ts
```diff
@@ -441,9 +441,13 @@ describe('Resource Packer: meta', () => {
...
 
   test('error during simulate', async () => {
-    await expect(externalClient.send.call({ method: 'error' })).rejects.toThrow(
-      'Error during resource population simulation in transaction 0',
-    )
+    try {
+      await externalClient.send.call({ method: 'error' })
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (e: any) {
+      expect(e.stack).toMatch(`err <--- Error`)
```
joe-p, line 448: Just noting that this was working before this change. Just wanted to make the test more explicit to catch any future regressions

neilcampbell (approved):
Nice one!

PR #357, opened 2024-12-16
feat: cover fees for app call transactions that send inner transactions

neilcampbell:
Add support for calculating fees for app call transactions that send inner transactions.


src/transaction/transaction.ts
```diff
@@ -310,279 +375,377 @@ async function getUnnamedAppCallResourcesAccessed(atc: algosdk.AtomicTransaction
...
+                return acc + -requiredFeeDelta
+              }
+              return acc
+            }, 0n),
+            new Map<number, bigint>(),
+          ] as const,
+        )
+    : [0n, new Map<number, bigint>()]
+
+  executionInfo.txns.forEach(({ unnamedResourcesAccessed: r }, i) => {
```
neilcampbell, line 456: The resource population code is mostly unaltered.

src/transaction/transaction.ts
```diff
@@ -310,279 +375,377 @@ async function getUnnamedAppCallResourcesAccessed(atc: algosdk.AtomicTransaction
...
+ * @param atc The ATC containing the txn group
+ * @param sendParams The send params for the transaction group
+ * @param executionContext Additional execution context used to determine how best to alter transactions in the group
+ * @returns A new ATC with the alterations applied
+ *
+ * @privateRemarks
+ * Parts of this function will eventually be implemented in algod. Namely:
+ * - Simulate will return information on how to populate reference arrays, see https://github.com/algorand/go-algorand/pull/6015
+ */
+async function alterGroupBasedOnSendParams(
```
neilcampbell, line 395: Should we expose this like we do with `populateAppCallResources`? It's possible that devs will want to calculate fees, populate etc without sending immediately.

Also open to suggestions for a better name.

If we did expose this, we could potentially deprecate `populateAppCallResources`, as it's quite likely devs will want to use these features together.

docs/capabilities/transaction-composer.md
```diff
@@ -33,17 +33,146 @@ The [methods to construct a transaction](../code/classes/types_composer.default.
...
+```
+
+## Sending a transaction
+
+Once you have constructed all the required transactions, they can be sent by calling `send()` on the `TransactionComposer`.
+Additionally `send()` takes a number of parameters which allow you to opt-in to some additional behaviours as part of sending the transaction or transaction group, mostly significantly `populateAppCallResources` and `coverAppCallInnerTransactionFees`.
+
+### populateAppCallResources
+
+As the name suggests, this setting automatically updates the relevant transactions in the group to include any resources required for the transactions to execute successfully. It leverages the simulate endpoint to discover the accessed resources, which have not been explicitly specified. This setting only applies when you have constucted at least one app call transaction.
```
robdmoore, line 55: Minor: might be worth giving some examples of what we mean by "resources" given that's quite a general word.

src/transaction/transaction.spec.ts
```diff
@@ -150,6 +153,516 @@ describe('transaction', () => {
...
+      const appFactory = algorand.client.getAppFactory({
+        appSpec: JSON.stringify(innerFeeContract),
+        defaultSender: testAccount,
+      })
+
+      appClient1 = (await appFactory.send.bare.create({ note: 'app1' })).appClient
+      appClient2 = (await appFactory.send.bare.create({ note: 'app2' })).appClient
+      appClient3 = (await appFactory.send.bare.create({ note: 'app3' })).appClient
+
+      appFactory.send.create
```
robdmoore, line 175: Is this meant to be here?

src/transaction/transaction.spec.ts
```diff
@@ -150,6 +153,516 @@ describe('transaction', () => {
...
+
+      await appClient1.fundAppAccount({ amount: algo(2) })
+      await appClient2.fundAppAccount({ amount: algo(2) })
+      await appClient3.fundAppAccount({ amount: algo(2) })
+    })
+
+    test('throws when no max fee is supplied', async () => {
```
robdmoore, line 182: Do we need maxFee to be required?

src/transaction/transaction.spec.ts
```diff
@@ -150,6 +153,516 @@ describe('transaction', () => {
...
+      const expectedFee = 7000n
+      const params = {
+        method: 'send_inners_with_fees',
+        args: [appClient2.appId, appClient3.appId, [0n, 0n, 0n, 0n, [0n, 0n]]],
+        staticFee: microAlgo(expectedFee - 1n),
+        coverAppCallInnerTransactionFees: true,
+      } satisfies Parameters<(typeof appClient1)['send']['call']>[0]
+
+      await expect(async () => await appClient1.send.call(params)).rejects.toThrow(
+        /Fees were too small to resolve execution info via simulate/,
```
robdmoore, line 242: I think this error should be a bit clearer.

src/types/app-deployer.ts
```diff
@@ -274,7 +274,7 @@ export class AppDeployer {
       } else {
         composer.addAppDelete({ appId: existingApp.appId, ...deleteParams })
       }
-      const result = await composer.send({ ...sendParams, suppressLog: true })
+      const result = await composer.send({ ...sendParams })
```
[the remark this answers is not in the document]
robdmoore: Hmm, maybe. Whats the default log level in normal dapps? This could mean we spew out new logs by default which may be confusing?

src/types/composer.ts
```diff
@@ -513,8 +527,10 @@ export class TransactionComposer {
   /** The ATC used to compose the group */
   private atc = new algosdk.AtomicTransactionComposer()
 
-  /** Map of txid to ABI method */
-  private txnMethodMap: Map<string, algosdk.ABIMethod> = new Map()
+  /** Map of transaction index in the atc to a max logical fee.
+   * This is set using the value of either maxFee or staticFee.
+   */
+  private txnMaxFees: Map<number, AlgoAmount> = new Map()
```
robdmoore, line 533: Checking: did you check all usages of the other similar `Map` to make sure you got all the places? There's quite a few code paths in this class and it's easy to miss stuff.

src/types/composer.ts
```diff
@@ -789,17 +805,23 @@ export class TransactionComposer {
...
     const group = atc.buildGroup()
 
     const txnWithSigners = group.map((ts, idx) => {
       // Remove underlying group ID from the transaction since it will be re-grouped when this TransactionComposer is built
       ts.txn.group = undefined
-      // If this was a method call stash the ABIMethod for later
+      // If this was a method call return the ABIMethod for later
       if (atc['methodCalls'].get(idx)) {
-        this.txnMethodMap.set(ts.txn.txID(), atc['methodCalls'].get(idx))
+        return {
```
robdmoore, line 816: Left to right this is differnet behaviour?

src/types/app-deployer.ts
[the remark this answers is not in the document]
neilcampbell: The default will log from debug level. verbose is not logged.
Worth noting that this only applies to the deploy method, so it's probably run quite infrequently inside a CD process.

src/transaction/transaction.spec.ts
neilcampbell: Good catch, will remove.

src/types/composer.ts
neilcampbell: I'll double check again, however I did check fairly extensively.
Worth noting that I haven't exposed the `coverAppCallInnerTransactionFees` option via the deprecated stateless functions to reduce the surface area.

src/types/composer.ts
neilcampbell: Now that we have an abstraction that can store this info alongside the algosdk.Transaction, we no longer need to stash it in state within the composer.

src/transaction/transaction.spec.ts
neilcampbell: Additionally, having this as a requirement means that we can use this value in the simulate transaction and we don't have to guess a value to use. If we guess too low the transaction can fail due to "fee too small", if we guess too high it can fail due to "account overspend".

src/transaction/transaction.spec.ts
neilcampbell: The full message is "Fees were too small to resolve execution info via simulate. You may need to increase an app call transaction maxFee.". Do you think this is clear enough?

src/types/composer.ts
```diff
@@ -1293,6 +1384,13 @@ export class TransactionComposer {
         suppressLog: params?.suppressLog,
         maxRoundsToWaitForConfirmation: waitRounds,
         populateAppCallResources: params?.populateAppCallResources,
+        coverAppCallInnerTransactionFees: params?.coverAppCallInnerTransactionFees,
+        executionContext: params?.coverAppCallInnerTransactionFees
```
robdmoore, line 1388: Is this check correct?

src/transaction/transaction.ts
robdmoore: Definitely useful as a building block people might want to use independently, but not a fan of this name, but don't have a better suggestion.

Random ideas:

* `prepareTransactionGroupForSending`
* `correctTransactionGroupForSending`

PR #358, opened 2024-12-16
fix: waitForIndexer waiting for transactions that failed

joe-p:
## Proposed Changes

- Fixes the bug that was uncovered in #356 (and probably has surfaced before)

A few questions:

1. Is there a reason why txids were used instead of round? Is there a concern a block will be in indexer but not a txn?
2. On cursory glance, this seems to be the only use-case for the transaction logger. Is there another usecase that I am missing or can we stop using this class and mark it for deprecation? (assuming it's not commonly used externally)

### DIFF
--- a/docs/code/classes/testing.TransactionLogger.md
+++ b/docs/code/classes/testing.TransactionLogger.md
@@ -130,14 +130,15 @@ ___
 
 ### waitForIndexer
 
-▸ **waitForIndexer**(`indexer`): `Promise`\<`void`\>
+▸ **waitForIndexer**(`algod`, `indexer`): `Promise`\<`void`\>
 
-Wait until all logged transactions IDs appear in the given `Indexer`.
+Wait until indexer has the last round from algod.
 
 #### Parameters
 
 | Name | Type |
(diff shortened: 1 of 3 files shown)


robdmoore:
Answered the first question at https://github.com/algorandfoundation/algokit-utils-ts/pull/356#issuecomment-2556286955

Re: the second question, it is useful to have a log of transaction IDs where you are testing code that performs multiple transactions and want to assert on the result of that code. The transaction logger provides a nice unobtrusive way to do that out of the box.

joe-p:
> Answered the first question at https://github.com/algorandfoundation/algokit-utils-ts/pull/356#issuecomment-2556286955

It's not clear to me why round would be slower. "wait for all transactions" (current implementation) and "wait for last round" (this PR) are the same question to the indexer, with the latter being more succinct and only requiring one call.  

> Re: the second question, it is useful to have a log of transaction IDs where you are testing code that performs multiple transactions and want to assert on the result of that code. The transaction logger provides a nice unobtrusive way to do that out of the box.

Are there any examples of how it'd be used?

PR #371, opened 2025-01-07
feat: registerErrorTransformer in composer

### LINKED ISSUE #368: App client simulate errors should be parsed into a LogicError
## Problem

When an error is thrown by simulate developers get the error with the PC but miss out on all the useful information `LogicError` provides. 

### Proposal

Add a way to turn errors during simulate into LogicErrors. Exact approach TBD

### Pros and Cons

TBD

## Dependencies

None

joe-p:
## Proposed Changes

- `registerErrorTransformer` is a new function on the composer that allow one to register functions that are used to transform errors

The main benefit is the improved error handling when composing together one group with multiple app clients. It also simplifies the error handling logic in the app client. When an `AppClient` is instantiated it registers the transformer to the `AlgorandClient`. This means errors for that app will always be properly handled, even when the method call did not come from the `AppClient`. These callbacks are also used on simulate, so this also closes #368. 

One thing I was going back and forth on is whether these functions should transform the error or if they are expected to raise the error themselves. The main reason I went with transforming is that transformers could add supplement information to the error, which seems useful.  

Keeping in draft until #365 is merged since I manually merged that in here


src/types/composer.ts
```diff
@@ -1363,12 +1392,19 @@ export class TransactionComposer {
...
 
+      for await (const cb of this.errorTransformers) {
+        const callbackResult = await cb(error)
+        if (callbackResult !== undefined) {
+          error = callbackResult
+        }
+      }
+
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       ;(error as any).simulateResponse = simulateResponse
```
tristanmenzel, line 1409: `Object.assign(error, { simulateResponse })` would remove the need for the `any` cast here.

src/types/composer.ts
```diff
@@ -505,6 +510,8 @@ export interface BuiltTransactions {
   signers: Map<number, algosdk.TransactionSigner>
 }
 
+export type ErrorTransformer<ErrorType> = (error: unknown) => Promise<ErrorType | undefined>
```
tristanmenzel, line 513: I don't think the generic type serves any value here? The signature could just be `(error: unknown) => Promise<unknown>`

src/types/composer.ts
```diff
@@ -1363,12 +1392,19 @@ export class TransactionComposer {
...
     if (failedGroup?.failureMessage) {
       const errorMessage = `Transaction failed at transaction(s) ${failedGroup.failedAt?.join(', ') || 'unknown'} in the group. ${failedGroup.failureMessage}`
-      const error = new Error(errorMessage)
+      let error = new Error(errorMessage)
 
       if (Config.debug) {
         await Config.events.emitAsync(EventType.TxnGroupSimulated, { simulateResponse })
       }
 
+      for await (const cb of this.errorTransformers) {
```
tristanmenzel, line 1401: Is the `await` required here? 🤔 `this.errorTransformers` is just a simple array, not an async generator

src/types/composer.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { beforeEach, describe, expect, test } from 'vitest'
+import { algorandFixture } from '../testing'
+
+const fixture = algorandFixture()
```
tristanmenzel, line 4: I think this should be declared inside the `describe` function or its lifetime will be that of the application

src/types/composer.ts
joe-p: My thinking was that some higher-level interface might require a transformer than returns a specific error type and in that case they could use the `ErrorTransformer` generic to express that. Given the simplicity of the signature it doesn't matter much because in that scenario that higher-level interface can just define their own type rather than use `ErrorTransformer` so I don't have a strong opinion either way.

src/types/algorand-client.ts
```diff
@@ -41,6 +40,9 @@ export class AlgorandClient implements AlgorandClientInterface {
     this._appDeployer = new AppDeployer(this._appManager, this._transactionSender, this._clientManager.indexerIfPresent)
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  errorTransformers: Map<number, ErrorTransformer<any>> = new Map()
```
tristanmenzel, line 44: Would it work to use a `Set<ErrorTransformer>` here and have deregistration work via reference instead of a returned id?

ie. 

```ts
const myTransformer = (e) => e
client.registerErrorTransformer(e)
client.registerErrorTransformer(e) // Automatic noop, as set will ignore duplicates
client.unregisterErrorTransformer(e) // e is removed from the set by reference
```

src/types/app-client.ts
```diff
@@ -1425,21 +1416,25 @@ export class AppClient {
...
-          .map((line) => `${(currentLine += 1)}: ${line}`)
-          .join('\n')
-        Config.logger.error(`${logicError.message}\n\n${stackWithLines}`)
-      }
-      throw logicError
+  private handleCallErrors = async (e: unknown) => {
+    // Only handle errors for this app.
+    // Because the error type is unknown, we turn it into a string to parse app ID
+    const appIdString = `app=${this._appId.toString()}`
+    if (!`${e}`.includes(appIdString)) {
```
tristanmenzel, line 1423: This logic feels a bit flaky. Can we test for a well known property instead? Ie. `if ('appId' in e) {` ?
Even better if it could be combined with an `instanceof` check.

src/types/app-factory-and-client.spec.ts
```diff
@@ -714,6 +714,28 @@ describe('ARC56: app-factory-and-app-client', () => {
...
+    try {
+      // Don't use the app client to call, but since we've instantiated one the error transformer should be registered
+      await appClient.algorand
+        .newGroup()
+        .addAppCallMethodCall({ appId: appClient.appId, method: appClient.getABIMethod('throwError')!, sender: testAccount })
+        .send()
+      invariant(false)
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (e: any) {
+      expect(JSON.stringify(e)).toMatch('this is an error')
```
tristanmenzel, line 735: An `instanceof` check would be better here, serializing unknown classes can result in infinite loops if two objects reference each other.

src/types/composer.ts
joe-p: Oh nice! I misunderstood how `for await of` worked. Removed

src/types/composer.ts
tristanmenzel: Yeah I say drop it. There's nothing to stop someone from having a strongly typed transformer and passing it in, but at this level we care nothing for the type, and actually dispose of it as soon as it's added to the `errorTransformers` collection

tristanmenzel (approved):
The general approach looks good to me 👍

src/types/app-client.ts
joe-p: Let me know what you think of [37c68b7](https://github.com/algorandfoundation/algokit-utils-ts/pull/371/commits/37c68b745be19e9ed4341357c644015e0cde6189)

src/types/algorand-client.ts
joe-p: ah yep much cleaner

src/types/app-client.ts
tristanmenzel: It's better, I guess I'm just not familiar with why `e.toString()` would reliably contain the string `app=X`, but not reliably have a property we could inspect instead?

joe-p: I addressed this in [c915e9c](https://github.com/algorandfoundation/algokit-utils-ts/pull/371/commits/c915e9ca9a098f81e14fd139eaaf2547c94a6e9b), but realized we could just go one step further and enforce that all input/output to transformers must be `Error` and just short-circuit if not. Keen to hear your thoughts on [f1fab2b](https://github.com/algorandfoundation/algokit-utils-ts/pull/371/commits/f1fab2b35a4809e02c90ee5912a7783a8f9b7091)

src/types/app-client.ts
```diff
@@ -1425,21 +1416,22 @@ export class AppClient {
...
-          ?.split('\n')
-          .map((line) => `${(currentLine += 1)}: ${line}`)
-          .join('\n')
-        Config.logger.error(`${logicError.message}\n\n${stackWithLines}`)
-      }
-      throw logicError
-    }
+  private handleCallErrors = async (e: Error) => {
+    // Only handle errors for this app.
+    const appIdString = `app=${this._appId.toString()}`
```
robdmoore, line 1421: How reliable is this?

src/types/composer.ts
```diff
@@ -1287,15 +1314,27 @@ export class TransactionComposer {
...
+          atc: this.atc,
+          suppressLog: params?.suppressLog,
+          maxRoundsToWaitForConfirmation: waitRounds,
+          populateAppCallResources: params?.populateAppCallResources,
+        },
+        this.algod,
+      )
+    } catch (e: unknown) {
+      let error = e
+      for (const cb of this.errorTransformers) {
```
robdmoore, line 1329: Minor: `cb`?

src/types/algorand-client-interface.ts
```diff
@@ -1,29 +1,5 @@
...
-    algod: Algodv2
-    indexer?: Indexer
-    network(): Promise<NetworkDetails>
-    isLocalNet(): Promise<boolean>
-    isTestNet(): Promise<boolean>
-    isMainNet(): Promise<boolean>
-  }
-}
+/** AlgorandClient interface used to eleminate circular dependencies */
+export type AlgorandClientInterface = InterfaceOf<AlgorandClient>
```
robdmoore, line 5: Why is this change in this PR?

src/types/app-client.ts
```diff
@@ -1170,29 +1171,29 @@ export class AppClient {
...
         return {
-          ...(await this.handleCallErrors(async () => this._algorand.send.appUpdate(await this.params.bare.update(params)))),
+          ...(await this._algorand.send.appUpdate(await this.params.bare.update(params))),
           ...(compiled as Partial<AppCompilationResult>),
         }
       },
       /** Signs and sends an opt-in call */
       optIn: (params?: AppClientBareCallParams & SendParams) => {
-        return this.handleCallErrors(() => this._algorand.send.appCall(this.params.bare.optIn(params)))
+        return this._algorand.send.appCall(this.params.bare.optIn(params))
```
robdmoore, line 1180: This is a nice side-effect of the change

robdmoore:
I like this, I think it's a neat solution.

My only worry is how reliable looking for `app={appid}` will be both now and in the future as avm changes...

I wonder if there is a way to pass in some of the original context or not so a structured comparison could be done rather than a magic string compare?

joe-p:
The problem is that since this is an error originating from an HTTP call we have no control over how granular the context we inject can be (i.e. we could inject group or the txid we waited for, but not per-transaction level details without also parsing the error) nor do we have control over the error itself. I don't think `appid=` is susceptible to being changed like the opcodes considering it is zero cost. While there is no guarantee it won't change, any potential change will be heavily considered. The ocpode change wouldn't have happened if it wasn't negatively impacting the networking, which `appid=` won't do.

src/types/app-client.ts
joe-p: Continuing this discussion in main thread for visibility

src/types/algorand-client-interface.ts
joe-p: I had to either update the interface in this PR and deal with merge conflicts when #365 is merged or merge #365 in this branch. Although as per my latest comment on #365 I'm not sure if this is the best approach. fwiw if I encountered having to do this before #365, I probably would have made the same change.

src/types/app-client.ts
joe-p: Yeah in general I want to look at features in the high level abstractions (i.e. generated client and app client) and see what we can do at a lower level to better support it and reduce the maintenance burden (and potential feature divergence) on these higher level classes.

src/types/composer.ts
```diff
@@ -1287,15 +1314,27 @@ export class TransactionComposer {
...
+          suppressLog: params?.suppressLog,
+          maxRoundsToWaitForConfirmation: waitRounds,
+          populateAppCallResources: params?.populateAppCallResources,
+        },
+        this.algod,
+      )
+    } catch (e: unknown) {
+      let error = e
+      for (const transformer of this.errorTransformers) {
+        if (!(error instanceof Error)) {
```
tristanmenzel, line 1330: I think this can be moved to before the loop, and just re-throw `e` if it's not an `Error`.

src/types/composer.ts
```diff
@@ -476,6 +476,15 @@ export type Txn =
   | { atc: algosdk.AtomicTransactionComposer; type: 'atc' }
   | ((AppCallMethodCall | AppCreateMethodCall | AppUpdateMethodCall) & { type: 'methodCall' })
 
+/**
+ * A function that transform an error into a new error.
+ *
+ * ErrorTransformers should be pessimistic about the input, hence the unknown type.
```
tristanmenzel, line 482:
```suggestion
```

src/types/algorand-client.ts
```diff
@@ -165,6 +183,7 @@ export class AlgorandClient implements AlgorandClientInterface {
       getSuggestedParams: () => this.getSuggestedParams(),
       defaultValidityWindow: this._defaultValidityWindow,
       appManager: this._appManager,
+      errorTransformers: [...this._errorTransformers.values()],
```
tristanmenzel, line 186: Shouldn't need the `.values()` here as a set is already [spreadable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Symbol.iterator)

src/types/composer.ts
joe-p: I intentionally wrote it in the loop because I want to be pessimistic about the return value of transformers. A bad transformer might return something that is not an error, thus breaking later transformers. Perhaps a better solution to this is to raise a specific `BadTransformer` error that wraps the original error and the object last returned

src/types/algorand-client.ts
joe-p: Good catch. Left over from when it used `Map`. Fixed in d8aff7e0028235b5da0176548feddfb42b4f188b

src/types/composer.ts
tristanmenzel: Yeah I think better not to hide a bad transformer.  Either assume they’re all adhering to the type or enforce it after each call.

joe-p: Yeah agreed. Addressed in [b75ff87](https://github.com/algorandfoundation/algokit-utils-ts/pull/371/commits/b75ff87aeacdd11cbde514cb65d7296e7850247a)

src/types/composer.ts
```diff
@@ -484,6 +484,12 @@ export type Txn =
  */
 export type ErrorTransformer = (error: Error) => Promise<Error>
 
+class BadTransformer extends Error {
+  constructor(originalError: unknown) {
```
tristanmenzel, line 488: The convention would be to pass the original error to the base type via the [cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) property.

Ie.

```ts
class BadTransformer extends Error {
  constructor(originalError: unknown) {
    super('An error transformer returned a non-error value.'', { cause: originalError} )
  }
}
```

tristanmenzel: Actually thinking more on this, maybe this is not an exact fit for using cause 🤔 

🤷 What do you think?

joe-p:
Now that #365 is merged this should be good to go. Only substantive change since last round of reviews was [68674c0](https://github.com/algorandfoundation/algokit-utils-ts/pull/371/commits/68674c02c7e83171bf00ad97a2cc51243b6f5994)

aorumbayev:
@joe-p do you consider this new functionality applicable to python utils as well? Given that there py version is up to speed with ts counterpart, makes sense to keep feature parity in mind when extending interfaces on either py or ts side.

joe-p:
Ah yes good point @aorumbayev. I could either PR into your branch next week or to reduce moving parts we can merge your PR first and then have a separate PR for this functionality. 

To avoid noise in the issue just DM on slack if you have a preference, otherwise I will make a PR into your branch next week

src/types/app-client.ts
```diff
@@ -504,6 +504,7 @@ export class AppClient {
     this._appSpec = AppClient.normaliseAppSpec(params.appSpec)
     this._appName = params.appName ?? this._appSpec.name
     this._algorand = params.algorand
+    this._algorand.registerErrorTransformer!(this.handleCallErrors)
```
tristanmenzel, line 507: Is the `!` necessary here?

joe-p: For now, yes. It can be removed next major release when we remove `AlgorandClientInterface`. See https://github.com/algorandfoundation/algokit-utils-ts/pull/365#issuecomment-2599791379

src/types/composer.ts
```diff
@@ -476,6 +476,26 @@ export type Txn =
   | { atc: algosdk.AtomicTransactionComposer; type: 'atc' }
   | ((AppCallMethodCall | AppCreateMethodCall | AppUpdateMethodCall) & { type: 'methodCall' })
 
+/**
+ * A function that transform an error into a new error.
```
tristanmenzel, line 480:
```suggestion
 * A function that transforms an error into a new error.
```

src/types/composer.ts
```diff
@@ -476,6 +476,26 @@ export type Txn =
...
+ *
+ * In most cases, an ErrorTransformer should first check if it can or should transform the error
+ * and return the input error if it cannot or should not transform it.
+ */
+export type ErrorTransformer = (error: Error) => Promise<Error>
+
+class NonErrorTransformer extends Error {
```
tristanmenzel, line 487:
```suggestion
class NonErrorTransformerError extends Error {
```

src/types/composer.ts
```diff
@@ -493,6 +513,11 @@ export type TransactionComposerParams = {
    * If not specified than an ephemeral one will be created.
```
tristanmenzel, line 513:
```suggestion
   * If not specified then an ephemeral one will be created.
```

src/types/instance-of.ts
```diff
@@ -0,0 +1,3 @@
+export type InterfaceOf<ClassType> = {
```
tristanmenzel, line 1: Could maybe benefit from a simple jsdoc comment since it appears to be included in the docs output above

aorumbayev:
> Ah yes good point @aorumbayev. I could either PR into your branch next week or to reduce moving parts we can merge your PR first and then have a separate PR for this functionality.
> 
> To avoid noise in this issue just DM on slack if you have a preference, otherwise I will make a PR into your branch next week

Unless there is some rush on this, might be more straightforward for you to add this after initial v3 prod release? There is still a bit of tweaks that may happen based on ongoing feedback session with DevRel. Would be significantly easier for you to introduce it once prod release is out and everything is settled. I think this should happen within 1-2 weeks (depending on how long feedback cycle takes), i can ping you as soon as prod version is out.

src/types/composer.ts
joe-p: I've renamed it to something a bit different, `InvalidErrorTransformerValue` which I think is a bit easier to understand

PR #372, opened 2025-01-16
feat: add heartbeat data to indexer types

neilcampbell:
Adding heartbeat transaction data to the indexer types.
This change is only required in v7 (hence `main-v7` target), as v8 leverages the indexer types introduced in algosdk@3.

### DIFF
--- a/src/types/indexer.ts
+++ b/src/types/indexer.ts
@@ -164,6 +164,8 @@ export interface TransactionResult extends Record<string, any> {
   'payment-transaction'?: PaymentTransactionResult
   /** If the transaction is a `stpf` transaction this will be populated see `tx-type` */
   'state-proof-transaction'?: StateProofTransactionResult
+  /** If the transaction is a `hb` transaction this will be populated see `tx-type` */
+  'heartbeat-transaction'?: HeartbeatTransactionResult
   /** [sgnr] this is included with signed transactions when the signing address does not equal the sender.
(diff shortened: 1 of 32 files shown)


src/types/indexer.ts
```diff
@@ -409,6 +411,52 @@ export interface StateProofTransactionResult {
...
+  'hb-address': string
+  /** [hbkd] HbKeyDilution must match HbAddress account's current KeyDilution. */
+  'hb-key-dilution': number
+  /** [hbprf] HbProof is a signature using HeartbeatAddress's partkey, thereby showing it is online. */
+  'hb-proof': {
+    /** [p] Public key of the heartbeat message.
+     *
+     * *Pattern:* `"^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==\|[A-Za-z0-9+/]{3}=)?$"`
+     */
+    'hb-pk'?: string
```
neilcampbell, line 426: This optionality is a little strange, however it mirrors the indexer api docs. https://developer.algorand.org/docs/rest-apis/indexer/#transactionheartbeat

lempira: I agree. All the props in `hb-proof` are optional, but `hb-proof` is not. Seems weird but good that it matches SDK and indexer API.

lempira (approved):

PR #382, opened 2025-02-26
feat: use algod's resoure population

joe-p:
## Proposed Changes

- Uses algod's resource population mechanism instead of doing it client side. 

Algod's algorithim is better tested and enables deterministic population (i.e. the same group always gets the same resources/order). 

Depends on https://github.com/algorand/js-algorand-sdk/pull/948, which depends on https://github.com/algorand/algorand-sdk-testing/pull/319, which depends on https://github.com/algorand/go-algorand/pull/6015


joe-p:
Noting that CI is currently failing because this only works when locally linking against the unreleased branch in this PR: https://github.com/algorand/js-algorand-sdk/pull/948

PR #387, opened 2025-03-21
feat!: replace AlgorandClientInterface with AlgorandClient

joe-p:
See the discussion in
https://github.com/algorandfoundation/algokit-utils-ts/pull/365 for context. Essentially by continuing to use the Interface any additional feature to AlgorandClient becomes a breaking change. The only reason for using the interface in the first place was for a rollup error, but this was a false positive and can be fixed by using the type import as shown in this commit

This PR should stay in draft until we can measure the impact to depdendent libraries and come up with a release strategy as we approach Algokit v3


lempira (approved):
Looks good

PR #390, opened 2025-04-11
fix: ensure getMap works properly with prefix

joe-p:
`getMap` was erroneously setting the prefix twice which resulted in the boxes not being found. 

*Note:* I originally planned to write the test contract with Algorand TypeScript, but it had a bug generating ARC56 state maps


tests/example-contracts/box_map/tsconfig.json
```diff
@@ -0,0 +1,103 @@
+{
```
lempira, line 1: Do we need all of the commented boilerplate configs for the example?

lempira (approved):
LGTM. Just a comment on tsconfig commented code.

tests/example-contracts/box_map/tsconfig.json
joe-p: Yeah good point. I reduced it to just what is recommended for TEALScript

PR #400, opened 2025-05-20
fix: properly identify newly created apps when handling errors

### LINKED ISSUE #398: AppFactory doesn't return custom error message when deploying
### Subject of the issue
When calling PuyaPy contract `@abimethod(create="require")` using `factory.deploy({ createParams: { ... } })`, custom error messages are not returned

### Your environment
AlgoKit: 2.7.0
Algokit-Utils": 9.1.0

### Steps to reproduce
Contract
```
@abimethod(create="require")
def create(self) -> None:
     assert False, "custom error message"
```

Deployment
```
factory.deploy({
     createParams: {
            sender: creator,
            method: "create",
            args: [],
      },
});
```

### Expected behaviour
You should see "custom error message"

### Actual behaviour
Generic error message with pc and opcodes

joe-p:
Uses the approval and clear programs to identify apps when app ID is zero. This allows us to properly parse error messages on creation since we historically used app ID to identify apps from error messages.

It should be noted that this only works for app created via an outer app call. Any inner creations will not be able to be mapped to app spec. https://github.com/algorand/go-algorand/issues/6329 is a feature request to add a hash of the program for this scenario so the app that failed can be identified

- Fixes #398


src/types/app-client.ts
```diff
@@ -1563,10 +1568,27 @@ export class AppClient {
...
+    // to identify if this is the correct app
+    if (this.appId === 0n) {
+      if (e.sentTransactions === undefined) return e
+
+      const txns = e.sentTransactions
+
+      const txn = txns.find((t) => e.message.includes(t.txID()))
+
+      if (
+        txn?.applicationCall?.approvalProgram.toString() !== this._lastCompiled?.approval?.toString() ||
```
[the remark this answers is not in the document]
joe-p: Resolved in [4f1d5d0](https://github.com/algorandfoundation/algokit-utils-ts/pull/400/commits/4f1d5d059a1d82ae6af997e26f352a0b150ac9a9)

src/types/app-client.ts
```diff
@@ -1563,10 +1568,38 @@ export class AppClient {
...
+    // to identify if this is the correct app
+    if (this.appId === 0n) {
+      if (e.sentTransactions === undefined) return e
+
+      const txns = e.sentTransactions
+
+      const txn = txns.find((t) => e.message.includes(t.txID()))
+
+      const programsEqual = (a: Uint8Array | undefined, b: Uint8Array | undefined) => {
+        if (a === undefined || b === undefined) return false
```
lempira, line 1582: If they are both undefined, `programEqual` would return false when they would technically be equal. Does that matter?

src/types/app-client.ts
```diff
@@ -1563,10 +1568,38 @@ export class AppClient {
...
+    // We can't use the app ID in an error to identify new apps, so instead we check the programs
+    // to identify if this is the correct app
+    if (this.appId === 0n) {
+      if (e.sentTransactions === undefined) return e
+
+      const txns = e.sentTransactions
+
+      const txn = txns.find((t) => e.message.includes(t.txID()))
+
+      const programsEqual = (a: Uint8Array | undefined, b: Uint8Array | undefined) => {
```
lempira, line 1581: could this fn be used elsewhere? Should it be a util?

src/types/app-factory-and-client.spec.ts
```diff
@@ -790,6 +791,16 @@ describe('ARC56: app-factory-and-app-client', () => {
     ).rejects.toThrow('custom error message')
   })
 
+  test('ARC56 error message on deploy', async () => {
+    const deployErrorFactory = localnet.algorand.client.getAppFactory({
+      // @ts-expect-error TODO: Fix this
```
lempira, line 796: what is this error? Should this be fixed now. If not, can it be ignored?

src/types/app-client.ts
joe-p: Undefined program means we can't properly check if this is the app we want. I've updated the name to make that more clear

src/types/app-client.ts
joe-p: Not that I can think of. If you prefer breaking it out into utils now we can but I don't think it needs to

lempira: ok. just checking if it could be used anywhere else.

lempira (approved):

PR #401, opened 2025-05-20
feat: support create method in AppClient

joe-p:
Main driver behind this change is allowing the factory to use an app client to reduce duplicated code

I've now realized I've deleted some methods that are public, so we might have to add those back in to prevent MAJOR version bump. Will keep in draft until that is done

### DIFF
--- a/src/types/app-client.ts
+++ b/src/types/app-client.ts
@@ -61,6 +61,7 @@ import { AppSpec, arc32ToArc56 } from './app-spec'
 import {
   AppCallMethodCall,
   AppCallParams,
+  AppCreateMethodCall,
   AppDeleteMethodCall,
   AppDeleteParams,
   AppMethodCall,
@@ -260,9 +261,9 @@ export interface FundAppAccountParams {
 /** Source maps for an Algorand app */
 export interface AppSourceMaps {
   /** The source map of the approval program */
-  approvalSourceMap: SourceMapExport
+  approvalSourceMap: algosdk.ProgramSourceMap
   /** The source map of the clear program */
-  clearSourceMap: SourceMapExport
+  clearSourceMap: algosdk.ProgramSourceMap
 }
 
 export interface SourceMapExport {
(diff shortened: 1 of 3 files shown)


src/types/app-client.ts
```diff
@@ -1222,6 +1239,9 @@ export class AppClient {
       update: async (params?: AppClientBareCallParams & AppClientCompilationParams) => {
         return this._algorand.createTransaction.appUpdate(await this.params.bare.update(params))
       },
+      create: async (params?: AppClientBareCallParams & AppClientCompilationParams) => {
+        return this._algorand.createTransaction.appCreate(await this.params.bare.update(params))
```
robdmoore, line 1243: Should that be create rather than update?

src/types/app-deployer.ts
```diff
@@ -300,9 +301,9 @@ export class AppDeployer {
       }
       const createIndex = await composer.count()
       if ('method' in deleteParams) {
-        composer.addAppDeleteMethodCall({ appId: existingApp.appId, ...deleteParams })
+        composer.addAppDeleteMethodCall({ ...deleteParams, appId: existingApp.appId })
       } else {
-        composer.addAppDelete({ appId: existingApp.appId, ...deleteParams })
```
robdmoore, line 305: Are these changes (in a few places) deliberate? In general I was pretty careful about ordering of spreading but it’s possible I got some wrong

aorumbayev:
Just my 2c, If any of these would change higher level interfaces exposed to end user, would just like to remind that utils py would have to be upgraded as well. 
Also would recommend checking the behaviour with client generators py|ts to see if any changes would need introduction into typed version of factory or client that get auto generated.

src/types/app-client.ts
```diff
@@ -260,9 +261,9 @@ export interface FundAppAccountParams {
 /** Source maps for an Algorand app */
 export interface AppSourceMaps {
   /** The source map of the approval program */
-  approvalSourceMap: SourceMapExport
+  approvalSourceMap: algosdk.ProgramSourceMap
```
neilcampbell, line 264: This would be extending reliance on algosdk types externally which I'm not sure we want to do.

joe-p: Ah good catch. Didn't realize the SDK used a class

PR #402, opened 2025-05-22
fix: make tealscript a dev dependency

PatrickDinh:
## Proposed Changes
Make tealscript a dev dependency because it's only used in the examples.


joe-p (approved):
Whoops must have forgot a dev flag at some point

PR #403, opened 2025-05-22
Suggested params

PatrickDinh:
## Proposed Changes

-
-
-


.npmrc
```diff
@@ -0,0 +1 @@
+@algorandfoundation:registry=https://npm.pkg.github.com
```
neilcampbell, line 1: This should be commented out, as we only need to use this when updating now.

neilcampbell (approved):

PR #407, opened 2025-05-27
feat: calculate fee with algokit core

PatrickDinh:
## Proposed Changes

-
-
-

### DIFF
--- a/src/types/algokit-core-bridge.ts
+++ b/src/types/algokit-core-bridge.ts
@@ -26,15 +26,10 @@ export function buildPayment(
     suggestedParams,
   }: algosdk.PaymentTransactionParams & algosdk.CommonTransactionParams,
 ) {
-  if (params.staticFee !== undefined) {
-    suggestedParams.fee = params.staticFee.microAlgo
-    suggestedParams.flatFee = true
-  }
-
   const txn: AlgokitCoreTransaction = {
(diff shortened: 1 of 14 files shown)


src/types/algokit-core-bridge.ts
```diff
@@ -1,23 +1,37 @@
...
+    receiver,
+    amount,
+    closeRemainderTo,
+    rekeyTo,
+    note,
+    lease,
+    suggestedParams,
+  }: algosdk.PaymentTransactionParams & algosdk.CommonTransactionParams,
+) {
+  if (params.staticFee !== undefined) {
```
neilcampbell, line 29: If we wanted to we could do away with using flatFee and changing the meaning of fee.

src/types/algokit-core-bridge.ts
```diff
@@ -1,26 +1,35 @@
 import * as algodApi from '@algorandfoundation/algokit-algod-api'
-import { addressFromString, Transaction as AlgokitCoreTransaction, encodeTransactionRaw } from '@algorandfoundation/algokit-transact'
+import {
+  addressFromString,
+  Transaction as AlgokitCoreTransaction,
```
neilcampbell, line 4:
```suggestion
  Transaction as AlgoKitCoreTransaction,
```

neilcampbell (approved):

Issue #363, opened 2024-12-20
TransactionLogger.waitForIndexer waits for failed transactions and transactions that have already been found


joe-p:
### Subject of the issue

There are currently two issues with `waitForIndexer`

1. `waitForIndexer` does not track which transactions have already been found, meaning it will waste time waiting for a transaction that is already known to be in indexer
2. `waitForIndexer` will wait (unscessfully) for failed transactions to show up indexer

There are two potential solutions

1. Use round instead of transaction ids in `waitForIndexer` (part of #358)
2. Pop transactions that have been found or failed (likely a separate array)

robdmoore:
Using latest round defeats the purpose of being scoped to the transactions in the logger, but what could be done is look up the transaction id(s) in the logger (the latest one is probably enough unless it's an invalid transaction?) and then wait for the round of that particular transaction.

joe-p:
Do we expect someone to use the transaction logger to send some transactions and then not use for others? I suppose it's just not clear to me why/how this would be used externally. Using transactions adds a lot of complexity vs simply asking "is indexer caught up" because we'd also need to have logic that handles failed transactions

robdmoore:
We can use "indexer caught up to algod" for sure, it's just slower if you are doing parallel test runs.

> Do we expect someone to use the transaction logger to send some transactions and then not use for others?

Not sure I understand this sentence? The logger isn't sending transactions just logging when they are sent.

joe-p:
Ah didn't think of parallel tests... I meant do we expect the logger to not have the latest transactions that were sent to algod, and in the case of parallel tests (where each test has its own instance of logger) then the answer could be yes. 

How much slower are we talking? Is the slowdown intermittent or is indexer consistently slow in catching up? I typically avoid using indexer so it's been awhile since I've had first hand experience with it. I'm just trying to gauge whether it's worth adding the complexity for tracking failed transactions

robdmoore:
The other option is do nothing. It's so unlikely with per test scoping that it's even an issue if we are only waiting for the last transaction.

robdmoore:
(this behaviour has been there since this functionality was first implemented and noone has complained about it except when it was waiting for all transactions with global scoping, which will naturally show the problem up)

robdmoore:
I agree in principle, but my point is just this is an edge case since most times you call this will be right after you have issued transactions you've waited for (and thus they will throw before you hit this method).

robdmoore:
[...]

PR #408, opened 2025-06-02
fix: return waitForIndexer when latest lastValid is reached

joe-p:
## Proposed Changes

Fix for #363

Basically avoids waiting for 20 seconds once it becomes impossible for the transaction to get indexed

### DIFF
--- a/src/testing/transaction-logger.ts
+++ b/src/testing/transaction-logger.ts
@@ -1,4 +1,5 @@
 import algosdk from 'algosdk'
+import { Config } from '../config'
 import { runWhenIndexerCaughtUp } from './indexer'
 import Algodv2 = algosdk.Algodv2
 import decodeSignedTransaction = algosdk.decodeSignedTransaction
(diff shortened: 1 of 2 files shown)


src/testing/transaction-logger.ts
```diff
@@ -53,7 +58,22 @@ export class TransactionLogger {
...
-    await runWhenIndexerCaughtUp(() => indexer.lookupTransactionByID(lastTxId).do())
+    await runWhenIndexerCaughtUp(async () => {
+      try {
+        await indexer.lookupTransactionByID(lastTxId).do()
+      } catch (e) {
+        // If the txid lookup failed, then try to look up the last valid round
+        // If that round exists, then we know indexer is caught up
+        if (this._latestLastValidRound) {
+          await indexer.lookupBlock(this._latestLastValidRound).do()
+          console.debug(
```
neilcampbell, line 69:
```suggestion
          Config.getLogger().debug(
```
Will also need the Config import

neilcampbell (approved):

PR #409, opened 2025-06-05
feat: send transaction group

PatrickDinh:
## Proposed Changes

-
-
-


src/algokit-core-bridge/transaction-composer.ts
```diff
@@ -0,0 +1,483 @@
...
+  /**
+   * Create a `TransactionComposer`.
+   * @param params The configuration for this composer
+   * @returns The `TransactionComposer` instance
+   */
+  constructor(params: TransactionComposerParams) {
+    this.algod = params.algodClient.algoKitCoreAlgod
+    this.algosdkAlgod = params.algodClient
+
+    const defaultGetTransactionParams = async (): Promise<TransactionParams> => {
```
neilcampbell, line 162: I don't believe this is ever used currently, which means we aren't getting transaction params using the core http client. If we want caching, then doing this could be a little tricky as the cache is stored in the AlgorandClient.

src/algokit-core-bridge/transaction-composer.ts
```diff
@@ -0,0 +1,483 @@
...
+      const responseContext = await callWithRetry(() => this.algod.rawTransactionResponse(new File(signedTxns, '')))
+      // Call handle response for error handling purposes only
+      await handleJSONResponse(responseContext, algosdk.modelsv2.PostTransactionsResponse)
+
+      // TODO: when app call is supported, this should be the txId of the first app call txn
+      const txIDToWaitFor = txIDs[0]
+
+      // TODO: can we use utils waitForConfirmation here?
+      // use core for waitForConfirmation?
+      await algosdk.waitForConfirmation(this.algosdkAlgod, txIDToWaitFor, params.maxRoundsToWaitForConfirmation ?? 5)
```
neilcampbell, line 408: Can we move this to the core http client?

src/algokit-core-bridge/transaction-composer.ts
```diff
@@ -0,0 +1,483 @@
...
+        Config.getLogger(params.suppressLog).verbose(`Group transaction (${groupId}) sent with ${transactionsToSend.length} transactions`)
+      } else {
+        Config.getLogger(params.suppressLog).verbose(
+          `Sent transaction ID ${txIDs[0]} ${transactionsToSend[0].transactionType} from ${transactionsToSend[0].sender.address}`,
+        )
+      }
+
+      let confirmations: algosdk.modelsv2.PendingTransactionResponse[] | undefined = undefined
+      if (!params.skipWaiting) {
+        confirmations = await Promise.all(txIDs.map(async (txID) => await this.algosdkAlgod.pendingTransactionInformation(txID).do()))
```
neilcampbell, line 420: Same for this?

neilcampbell (approved):

PR #427, opened 2025-08-28
Feat/core abi

PatrickDinh:
## Proposed Changes

-
-
-


src/core/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+  CHECKSUM_ADDRESS_ERROR_MSG,
+  MALFORMED_ADDRESS_ERROR_MSG,
+} from './constants'
+import * as nacl from './nacl/naclWrappers'
+
+/**
+ * Check whether the environment is ReactNative
+ * @returns True if ReactNative, false otherwise
+ */
+export function isReactNative() {
```
neilcampbell, line 15: If we remove the tweetnacl dependency then we no longer need this which is nice.

src/core/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+export function bytesToBigInt(bytes: Uint8Array) {
+  let res = BigInt(0)
+  const buf = new DataView(bytes.buffer, bytes.byteOffset)
+  for (let i = 0; i < bytes.length; i++) {
+    res = BigInt(Number(buf.getUint8(i))) + res * BigInt(256)
+  }
+  return res
+}
+
+// TODO: this will be replaced by the real logic when this branch is merged
```
neilcampbell, line 77: This logic I have in an address.ts

src/core/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+ */
+export function isReactNative() {
+  const { navigator } = globalThis as { navigator?: { product?: string } }
+  if (typeof navigator === 'object' && navigator.product === 'ReactNative') {
+    return true
+  }
+  return false
+}
+
+export function arrayEqual<T>(a: ArrayLike<T>, b: ArrayLike<T>): boolean {
```
neilcampbell, line 23: Maybe we could extract into arrays.ts so we have more specialised helpers?

src/core/utils.ts
```diff
@@ -0,0 +1,110 @@
...
+  return c
+}
+
+/**
+ * bigIntToBytes converts a BigInt to a big-endian Uint8Array for encoding.
+ * @param bi - The bigint to convert.
+ * @param size - The size of the resulting byte array.
+ * @returns A byte array containing the big-endian encoding of the input bigint
+ */
+export function bigIntToBytes(bi: bigint | number, size: number) {
```
neilcampbell, line 49: I think these bigint to bytes are potentially abi specific, so you could potentially move into the abi namespace into a bigint.ts file?

src/core/constants.ts
```diff
@@ -0,0 +1,12 @@
+export const STATIC_ARRAY_REGEX = /^(.+)\[(\d+)\]$/
+export const UFIXED_REGEX = /^ufixed(\d+)x(\d+)$/
+
+export const MALFORMED_ADDRESS_ERROR_MSG = 'address seems to be malformed'
```
neilcampbell, line 4: We'll have some conflicts on this file. I've generally opted to remove `ALGORAND_` prefix, as it feels a little redundant.

src/core/abi/errors.ts
```diff
@@ -0,0 +1,20 @@
+export class ValidationError extends Error {
```
neilcampbell, line 1: I'm assuming these error classes are to match the rust behaviour?
There certainly a difference in how the errors are unwrapped, so might be worth a quick chat.

neilcampbell (commented):
The index.ts doesn't have any exports.
I was thinking this file should contain all the exports matching the FFI bound language equivalent, which would allow us to visually check consistency.

neilcampbell (approved):

PR #431, opened 2025-09-07
feat: add core composer support

neilcampbell:


src/core/utils/transactions/common.ts
```diff
@@ -0,0 +1,205 @@
...
+  NonParticipationKeyRegistrationComposerTransaction,
+  OfflineKeyRegistrationComposerTransaction,
+  OnlineKeyRegistrationComposerTransaction,
+} from './key-registration'
+import { AccountCloseComposerTransaction, PaymentComposerTransaction } from './payment'
+
+export type TransactionComposerTransaction = { type: ComposerTransactionType.Transaction; data: Transaction }
+export type TransactionWithSignerComposerTransaction = { type: ComposerTransactionType.TransactionWithSigner; data: TransactionWithSigner }
+
+export type TransactionHeader = {
```
PatrickDinh, line 30: I'm wondering whether we have the TransactionHeader type or not. Currently it exists in `core` but the plan is we will replace them with macro soon. Should we build TS without this type so that we don't have to do refactor in the near future?

src/core/utils/transactions/composer.ts
```diff
@@ -0,0 +1,1406 @@
...
+      // TransactionPlaceholder or DefaultValuePlaceholder
+      continue
+    } else if (isTransactionArg(arg)) {
+      composerTransactions.push({ type: ComposerTransactionType.Transaction, data: arg })
+      continue
+    } else if (isTransactionWithSignerArg(arg)) {
+      composerTransactions.push({ type: ComposerTransactionType.TransactionWithSigner, data: arg })
+      continue
+    } else if (isComposerTransactionParamsArg(arg)) {
+      composerTransactions.push(arg)
```
PatrickDinh, line 854: Looks like this method doesn't handle nested method call args

src/core/utils/transactions/common.ts
neilcampbell: Yeah I was tossing up wether we do or don't. I might keep like this a defer to the PR we add support for heartbeat and stateproof transactions.

src/core/utils/transactions/composer.ts
neilcampbell: Yeah I think you might be right, let me fix that.

PR #443, opened 2025-10-01
feat: deterministic resource population

joe-p:
This PR ensures that resource population is deterministic. It should be noted that there is no way to guarantee that the algorithm we have is the most efficient and what works best for one group may not be best for another. For now, having determinism is a good starting point to eliminate edge-case failures. In the future we could potentially add retry mechanism with different ordering.


src/transaction/transaction.ts
```diff
@@ -329,8 +329,34 @@ async function getGroupExecutionInfo(
...
+      const bStr = `${b.app}-${b.name}`
+      return compare(aStr, bStr)
+    })
+    sortedResources.appLocals?.sort((a, b) => {
+      const aStr = `${a.app}-${a.account}`
+      const bStr = `${b.app}-${b.account}`
+      return compare(aStr, bStr)
+    })
+    sortedResources.assetHoldings?.sort((a, b) => {
+      const aStr = `${a.asset}-${b.account}`
```
YaroslavNekryach, line 352:
```suggestion
      const aStr = `${a.asset}-${a.account}`
```

joe-p: Thanks!

lempira (approved):

src/transaction/transaction.spec.ts
```diff
@@ -1138,6 +1138,88 @@ describe('Resource population: meta', () => {
...
     expect(boxRef?.appIndex).toBe(0n)
   })
+
+  test('order is deterministic', async () => {
+    const { algorand, context } = fixture
+
+    const testAccount = context.testAccount
+
+    const v9AppFactory = algorand.client.getAppFactory({
+      appSpec: JSON.stringify(v9ARC32),
```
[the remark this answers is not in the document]
joe-p: It's not needed because v8 is just txn-specific resources, thus we don't have any decisions to make regarding the order they are populated

PR #449, opened 2025-10-30
Feat/transact integration

PatrickDinh:
## Proposed Changes

-
-
-

### DIFF
--- a/src/transaction/transaction.ts
+++ b/src/transaction/transaction.ts
@@ -518,6 +518,7 @@ export async function prepareGroupForSending(
       }
 
       if (r && !hasAccessReferences) {
+        if (r.boxes || r.extraBoxRefs) throw Error('Unexpected boxes at the transaction level')
         if (r.appLocals) throw Error('Unexpected app local at the transaction level')
         if (r.assetHoldings)
           throw Error('Unexpected asset holding at the transaction level')
@@ -527,10 +528,7 @@ export async function prepareGroupForSending(
           accountReferences: [...(group[i].txn?.appCall?.accountReferences ?? []), ...(r.accounts ?? [])],
           appReferences: [...(group[i].txn?.appCall?.appReferences ?? []), ...(r.apps ?? [])],
           assetReferences: [...(group[i].txn?.appCall?.assetReferences ?? []), ...(r.assets ?? [])],
-          boxReferences: [
-            ...(group[i].txn?.appCall?.boxReferences ?? []),
-            ...(r.boxes?.map((b) => ({ appId: b.app, name: b.name })) ?? []),
-          ],
+          boxReferences: [...(group[i].txn?.appCall?.boxReferences ?? [])],
         } satisfies Partial<ApplicationTransactionFields>
 
         const accounts = group[i].txn.appCall?.accountReferences?.length ?? 0
@@ -925,6 +923,9 @@ export const sendAtomicTransactionComposer = async function (atcSend: AtomicTran
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const err = new Error(errorMessage) as any
     err.cause = e
+    if (typeof e === 'object') {
+      err.name = e.name
+    }
 
     if (Config.debug && typeof e === 'object') {
       err.traces = []
--- a/src/types/__snapshots__/app-client.spec.ts.snap
+++ b/src/types/__snapshots__/app-client.spec.ts.snap
@@ -5,5 +5,5 @@ exports[`application-client > Errors > Display nice error messages when there is
 INFO: App TestingApp not found in apps created by ACCOUNT_1; deploying app with version 1.0.
 VERBOSE: Sent transaction ID TXID_1 appl from ACCOUNT_1
 DEBUG: App created by ACCOUNT_1 with ID APP_1 via transaction TXID_1
-ERROR: Received error executing Atomic Transaction Composer and debug flag enabled; attempting simulation to get more information | [{"cause":{},"traces":[]}]"
+ERROR: Received error executing Atomic Transaction Composer and debug flag enabled; attempting simulation to get more information | [{"cause":{},"name":"Error","traces":[]}]"
 `;
--- a/src/types/asset-manager.ts
+++ b/src/types/asset-manager.ts
@@ -166,7 +166,7 @@ export class AssetManager {
    * @returns The asset information
    */
   public async getById(assetId: bigint): Promise<AssetInformation> {
-    const asset = await this._algod.getAssetById(Number(assetId))
+    const asset = await this._algod.getAssetById(assetId)
 
     return {
       assetId: BigInt(asset.index),
@@ -204,7 +204,7 @@ export class AssetManager {
    * @returns The account asset holding information
    */
   public async getAccountInformation(sender: string | Address, assetId: bigint): Promise<AccountAssetInformation> {
-    const info = await this._algod.accountAssetInformation(sender.toString(), Number(assetId))
+    const info = await this._algod.accountAssetInformation(sender.toString(), assetId)
 
     return {
       assetId: BigInt(assetId),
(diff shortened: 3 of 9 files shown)


packages/transact/src/transactions/app-call.ts
```diff
@@ -86,26 +86,26 @@ export type AppCallTransactionFields = {
    * List of accounts in addition to the sender that may be accessed
    * from the app's approval program and clear state program.
    */
-  accountReferences?: string[]
+  accounts?: string[]
```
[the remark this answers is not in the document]
PatrickDinh: After reviewing I think that we still should make the changes. `accounts`, `foreignApps`, `foreignAssets` and `boxes` are terms widely used by Algorand docs. For example, ARC-4 [docs](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#reference-types)
and [here](https://github.com/algorandfoundation/ARCs/blob/main/ARCs/arc-0004.md#reference-types)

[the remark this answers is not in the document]
PatrickDinh: if we go with the new names, I'll update the validation error messages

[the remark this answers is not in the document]
PatrickDinh: I reverted it to keep the field names consistent with the current version of utils

packages/transact/src/transactions/signed-transaction.ts
```diff
@@ -10,7 +10,7 @@ export type SignedTransaction = {
   /**
    * The transaction that has been signed.
    */
-  transaction: Transaction
+  txn: Transaction
```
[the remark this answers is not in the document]
PatrickDinh: I renamed this because the `SignedTransaction` type is used within `PendingTransactionResponse` as 
```
  /**
   * The raw signed transaction.
   */
  txn: SignedTransaction
```
It would be strange to have to `response.txn.transaction` to access the transaction. Therefore, I renamed this to match with the current sdk. The user still can access the transaction by `response.txn.txn`.

src/transaction/transaction.ts
```diff
@@ -941,7 +943,7 @@ export const sendAtomicTransactionComposer = async function (atcSend: AtomicTran
       if (simulate && simulate.txnGroups[0].failedAt) {
         for (const txn of simulate.txnGroups[0].txnResults) {
           err.traces.push({
-            trace: txn.execTrace?.toEncodingData(),
+            trace: undefined, // TODO: need to encode txn.execTrace?.toEncodingData(), SimulationTransactionExecTrace
```
PatrickDinh: This will be resolved in the next PR

src/transaction/transaction.ts
```diff
@@ -891,38 +898,33 @@ export const sendAtomicTransactionComposer = async function (atcSend: AtomicTran
...
+      groupId: groupId!,
+      confirmations: (confirmations ?? []).map(wrapPendingTransactionResponse),
+      txIds: transactionsToSend.map((t) => getTransactionId(t)),
+      transactions: transactionsToSend.map((t) => new TransactionWrapper(t)),
       returns: result.methodResults.map((r, i) => getABIReturnValue(r, methodCalls[i]!.returns.type)),
-    } as SendAtomicTransactionComposerResults
+    } satisfies SendAtomicTransactionComposerResults
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
   } catch (e: any) {
+    // TODO: PD - look into error handling here again, it's possible that we don't need this comment anymore
```
PatrickDinh: will resolve in the composer PR

src/transaction/transaction.ts
```diff
@@ -491,41 +502,43 @@ export async function prepareGroupForSending(
...
-    if (sendParams.populateAppCallResources && group[i].txn.type === TransactionType.appl) {
+    if (sendParams.populateAppCallResources && group[i].txn.type === TransactionType.AppCall) {
       const hasAccessReferences = appCallHasAccessReferences(group[i].txn)
 
       if (hasAccessReferences && (r || executionInfo.groupUnnamedResourcesAccessed)) {
         indexesWithAccessReferences.push(i)
       }
 
       if (r && !hasAccessReferences) {
-        if (r.boxes || r.extraBoxRefs) throw Error('Unexpected boxes at the transaction level')
```
neilcampbell: Is this checked elsewhere, hence is not required?

src/types/__snapshots__/app-client.spec.ts.snap
```diff
@@ -5,5 +5,5 @@ exports[`application-client > Errors > Display nice error messages when there is
 INFO: App TestingApp not found in apps created by ACCOUNT_1; deploying app with version 1.0.
 VERBOSE: Sent transaction ID TXID_1 appl from ACCOUNT_1
 DEBUG: App created by ACCOUNT_1 with ID APP_1 via transaction TXID_1
-ERROR: Received error executing Atomic Transaction Composer and debug flag enabled; attempting simulation to get more information | [{"cause":{},"name":"Error","traces":[]}]"
+ERROR: Received error executing Atomic Transaction Composer and debug flag enabled; attempting simulation to get more information | [{"cause":{},"traces":[]}]"
```
neilcampbell: I don't understand why the name is missing?

src/types/algo-http-client-with-retry.ts
```diff
@@ -1,5 +1,5 @@
-import { IntDecoding, parseJSON, stringifyJSON } from 'algosdk'
-import { BaseHTTPClientResponse, Query, URLTokenBaseHTTPClient } from 'algosdk/client'
+import { IntDecoding, parseJSON, stringifyJSON } from '@algorandfoundation/sdk'
+import { BaseHTTPClientResponse, Query, URLTokenBaseHTTPClient } from '@algorandfoundation/sdk'
```
neilcampbell: Given our new clients don't use `URLTokenBaseHTTPClient` how does the retry mechanism work? Also we have a skipped Retry test (in client-manager.spec.ts) which we can use to test this behaviour.

src/types/asset-manager.ts
```diff
@@ -203,7 +204,7 @@ export class AssetManager {
    * @returns The account asset holding information
    */
   public async getAccountInformation(sender: string | Address, assetId: bigint): Promise<AccountAssetInformation> {
-    const info = await this._algod.accountAssetInformation(sender, Number(assetId)).do()
+    const info = await this._algod.accountAssetInformation(sender.toString(), Number(assetId))
```
neilcampbell:
```suggestion
    const info = await this._algod.accountAssetInformation(sender.toString(), assetId)
```

src/transaction/transaction.ts
PatrickDinh: AI slop 😢 my bad for didn't pick this one up. It was trying to fix a compile issues below
```
          boxes: [...(group[i].txn?.applicationCall?.boxes ?? []), ...(r.boxes ?? [])],
```
In this code, `r.boxes` is `never` because of this `if` statement

src/types/algo-http-client-with-retry.ts
PatrickDinh: The retry logic was moved into the `algod_client`
https://github.com/algorandfoundation/algokit-utils-ts/blob/8864a3eaf9979d4b3d82f3f08bdbc79c384e5bc9/packages/algod_client/src/core/fetch-http-request.ts#L70
I think that we should:
- Update `algod_client` to support a mechanism to log messages. 👈 this is to match with the existing behavior, Currently, when the request failed and retried, there was no log.
- Move the retry test to `algod_client`

src/types/__snapshots__/app-client.spec.ts.snap
PatrickDinh: I reverted the error handling logic back to restore the name. Originally I was planning to do it later and have the name field for our custom ApiError only because honestly I think `name: Error` is a generic and not needed.

tests/example-contracts/client/TestContractClient.ts
```diff
@@ -210,23 +212,23 @@ export const APP_SPEC: AppSpec = {
 /**
  * Defines an onCompletionAction of 'no_op'
  */
-export type OnCompleteNoOp = { onCompleteAction?: 'no_op' | OnApplicationComplete.NoOpOC }
+export type OnCompleteNoOp = { onCompleteAction?: 'no_op' | OnApplicationComplete.NoOp }
```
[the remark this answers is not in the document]
PatrickDinh: I was confused here, turns out the version in algosdk has the `OC` suffix. I thought that the suffix was added by utils layer.

neilcampbell (approved):

PR #454, opened 2025-11-10
chore: additional algod client refinements

neilcampbell:
- Lots of algod client refinements and testing
- Adding lmsig support
- Refactoring the API encoding/decode logic to work with transactions

Below are the main naming differences between JS algosdk (left) and the new algod client (right), with the new algod client providing more consistency in naming and aligning to the defined OAS spec operation id.
`versionsCheck()` -> `getVersion()`
`block()` -> `getBlock()`
`getAssetByID()` -> `getAssetById()`
`getApplicationByID()` -> `getApplicationById()`
`status()` -> `getStatus()`
`supply()` -> `getSupply()`
`compile()` -> `tealCompile()`
`disassemble()` -> `tealDisassemble()`
`dryrun()` -> `tealDryrun()`
`genesis()` -> `getGenesis()`
`pendingTransactionsInformation()` -> `getPendingTransactions()`
`pendingTransactionByAddress()` -> `getPendingTransactionsByAddress()`
`getBlockOffsetTimestamp()` -> `getBlockTimeStampOffset()`
`setBlockOffsetTimestamp()` -> `setBlockTimeStampOffset()`
`ready()` -> `getReady()`


packages/transact/src/transactions/signed-transaction.ts
```diff
@@ -95,11 +94,18 @@ export type LogicSignature = {
...
   /**
-   * Multisig for delegated logic sig
+   * Legacy multisig for delegated logic sig
    */
   multiSignature?: MultisigSignature
+
+  /**
+   * Multisig for delegated logic sig
+   */
+  logicMultiSignature?: MultisigSignature
```
neilcampbell: This field in algod is called `lmsig`. I'm assuming it stands for this, however I couldn't get confirmation from looking through the algod codebase.

Alternatively here, we could change to use lmsig and msig.

packages/algod_client/src/core/fetch-http-request.ts
```diff
@@ -101,15 +91,15 @@ export class FetchHttpRequest extends BaseHttpRequest {
 
     const status = extractStatus(error)
     if (status !== undefined) {
-      const retryStatuses = this.config.retryStatusCodes ?? RETRY_STATUS_CODES
+      const retryStatuses = RETRY_STATUS_CODES
       if (retryStatuses.includes(status)) {
```
PatrickDinh: nitpick: maybe we can do this
```suggestion
      if (RETRY_STATUS_CODES.includes(status)) {
```

packages/common/src/codecs/models/array-model.spec.ts
```diff
@@ -0,0 +1,400 @@
...
+      describe('default values', () => {
+        test.each<{ value: number[] | undefined | null; description: string }>([
+          { value: [], description: 'empty array (default value)' },
+          { value: undefined, description: 'undefined' },
+          { value: null, description: 'null' },
+        ])('should encode $description to undefined', ({ value }) => {
+          expect(codec.encode(value, 'json')).toBeUndefined()
+          expect(codec.encode(value, 'msgpack')).toBeUndefined()
+        })
+      })
```
PatrickDinh: this test confuses me because the doc for `encode` is `Encode a value, always returning the value regardless of if it is default` but this test expects an `undefined`

packages/indexer_client/src/apis/api.service.ts
```diff
@@ -471,17 +404,17 @@ export class IndexerApi {
       afterTime?: string
       currencyGreaterThan?: number | bigint
       currencyLessThan?: number | bigint
-      address?: string
+      address?: string | Address
```
PatrickDinh: should we use `ReabableAddress` here?

packages/transact/src/transactions/signed-transaction.ts
PatrickDinh: From this PR https://github.com/algorand/js-algorand-sdk/pull/999 it stands for `logic signature multisignature`. I think the name `logicMultiSignature` is ok

packages/common/src/codecs/composite/array.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+  addressArrayCodec,
+  bigIntArrayCodec,
+  booleanArrayCodec,
+  bytesArrayCodec,
+  numberArrayCodec,
+  stringArrayCodec,
+} from './array'
+
+describe('ArrayCodec', () => {
+  const testNumberArrayCodec = new ArrayCodec(numberCodec)
```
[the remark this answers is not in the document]
neilcampbell: This was to ensure we could construct a new variant.

packages/indexer_client/src/apis/api.service.ts
neilcampbell: This has been done in the indexer PR.

PR #457, opened 2025-11-12
Composer integration

PatrickDinh:
## Proposed Changes

-
-
-

### DIFF
--- a/src/transaction/transaction.ts
+++ b/src/transaction/transaction.ts
@@ -92,7 +92,7 @@ export async function populateAppCallResources(composer: TransactionComposer) {
  *
  * @param composer The Transaction Composer containing the txn group
  * @param sendParams The send params for the transaction group
- * @param additionalAtcContext Additional context used to determine how best to change the transactions in the group
+ * @param additionalContext Additional context used to determine how best to change the transactions in the group
  * @returns A new Transaction Composer with the changes applied
  *
  * @privateRemarks
@@ -102,15 +102,15 @@ export async function populateAppCallResources(composer: TransactionComposer) {
 export async function prepareGroupForSending(
   composer: TransactionComposer,
   sendParams: SendParams,
-  additionalAtcContext?: AdditionalTransactionComposerContext,
+  additionalContext?: AdditionalTransactionComposerContext,
 ) {
   const newComposer = composer.clone({
     coverAppCallInnerTransactionFees: sendParams.coverAppCallInnerTransactionFees ?? false,
     populateAppCallResources: sendParams.populateAppCallResources ?? true,
   })
 
-  if (additionalAtcContext?.maxFees) {
-    newComposer.setMaxFees(additionalAtcContext?.maxFees)
+  if (additionalContext?.maxFees) {
+    newComposer.setMaxFees(additionalContext?.maxFees)
   }
 
   await newComposer.build()
--- a/docs/code/modules/index.md
+++ b/docs/code/modules/index.md
@@ -383,7 +383,7 @@ ___
 
 ### prepareGroupForSending
 
-▸ **prepareGroupForSending**(`composer`, `sendParams`, `additionalAtcContext?`): `Promise`\<[`TransactionComposer`](../classes/types_composer.TransactionComposer.md)\>
+▸ **prepareGroupForSending**(`composer`, `sendParams`, `additionalContext?`): `Promise`\<[`TransactionComposer`](../classes/types_composer.TransactionComposer.md)\>
 
 #### Parameters
 
@@ -391,7 +391,7 @@ ___
 | :------ | :------ | :------ |
 | `composer` | [`TransactionComposer`](../classes/types_composer.TransactionComposer.md) | The Transaction Composer containing the txn group |
 | `sendParams` | [`SendParams`](../interfaces/types_transaction.SendParams.md) | The send params for the transaction group |
-| `additionalAtcContext?` | [`AdditionalTransactionComposerContext`](../interfaces/types_transaction.AdditionalTransactionComposerContext.md) | Additional context used to determine how best to change the transactions in the group |
+| `additionalContext?` | [`AdditionalTransactionComposerContext`](../interfaces/types_transaction.AdditionalTransactionComposerContext.md) | Additional context used to determine how best to change the transactions in the group |
 
 #### Returns
 


packages/transact/src/encoding/transaction-dto.ts
```diff
@@ -122,6 +122,9 @@ export type TransactionDto = {
   /** Extra program pages */
   apep?: number
 
+  /**  Reject version */
+  aprv?: bigint | number
```
neilcampbell:
```suggestion
  aprv?: number
```

src/transaction/perform-transaction-composer-simulate.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Performs a simulation of the transactions loaded into the given TransactionComposer.
+ * Uses empty transaction signers for all transactions.
+ *
+ * @param composer The TransactionComposer with transaction(s) loaded.
+ * @returns The simulation result, which includes various details about how the transactions would be processed.
+ */
+export async function performTransactionComposerSimulate(composer: TransactionComposer, options?: RawSimulateOptions) {
+  // NOTE: the existing version takes atc as params, then call atc.buildGroup to get the transactions
+  // The state of the atc is unknown, whether it has resource populated or not
```
neilcampbell: Resource population in the old composer only ran on send, so it's likely safe to assume that it was not run.

src/transaction/perform-transaction-composer-simulate.ts
```diff
@@ -0,0 +1,24 @@
...
+  // In this version, we call composer.simulate which doesn't do resource population
```
neilcampbell: I would have thought that `build()` would need to be called and as a result it would have been ran? I don't think we want to run resource pop on simulate though, as it essentially results in 2 simulates.

src/transaction/transaction.ts
```diff
@@ -425,547 +278,117 @@ async function getGroupExecutionInfo(
...
-            // account is available as an app account
-            t.txn.appCall?.appReferences?.map((a) => algosdk.getApplicationAddress(a).toString()).includes(account.toString()) ||
-            // account is available since it's in one of the fields
-            Object.values(t.txn).some((f) =>
-              stringifyJSON(f, (_, v) => (v instanceof Address ? v.toString() : v))?.includes(account.toString()),
-            )
-          )
+  transactionsWithSigners.forEach((txnWithSigner, index) => {
+    if (txnWithSigner.txn.type !== TransactionType.AppCall) {
+      newComposer.addTransaction(txnWithSigner.txn)
```
neilcampbell: Do we also want to add the signer?

src/transactions/clone.ts
```diff
@@ -0,0 +1,150 @@
...
+      }
+    }
+    return cloned
+  }
+
+  // Fallback - return as-is
+  return value
+}
+
+export function deepCloneTransactionParams<T extends Record<string, any>>(params: T): T {
```
neilcampbell: What is this used for?

src/types/app-client.ts
```diff
@@ -2162,24 +2160,31 @@ export class ApplicationClient {
...
       // The method is readonly
-      this.appSpec.hints[this.getABIMethodSignature(this.getABIMethod(call.method)!)].read_only
+      this.appSpec.hints?.[this.getABIMethodSignature(this.getABIMethod(call.method)!)]?.read_only
     ) {
-      const atc = new AtomicTransactionComposer()
-      await this.callOfType({ ...call, sendParams: { ...call.sendParams, atc } }, 'no_op')
-      const result = await atc.simulate(this.algod)
+      const transactionComposer = new TransactionComposer({
+        algod: this.algod,
+        getSigner: (address) => {
```
neilcampbell: We do this pattern in a few places, so we could extract this function and share across the code.

src/types/composer.ts
```diff
@@ -615,9 +302,164 @@ export class TransactionComposer {
     this.getSuggestedParams = params.getSuggestedParams ?? defaultGetSuggestedParams
     this.getSigner = params.getSigner
     this.defaultValidityWindow = params.defaultValidityWindow ?? this.defaultValidityWindow
```
neilcampbell: Why was `defaultValidityWindowIsExplicit` deleted?

src/types/composer.ts
```diff
@@ -615,9 +302,164 @@ export class TransactionComposer {
...
+  }
+
+  /**
+   * Helper function to clone a single transaction
+   * @param txn The transaction to clone
+   * @returns The cloned transaction
+   */
+  private cloneTransaction(txn: Txn): Txn {
+    switch (txn.type) {
+      case 'pay':
```
neilcampbell: Can you do something like this:
```ts
  switch (txn.type) {
      case 'pay':
      case 'assetCreate':
      case 'assetConfig':
      // All the other granular transaction types we define
        return {
          type: txn.type,
          data: deepCloneTransactionParams(txn.data),
        }
        // Other case statements
    }
  }
```

src/types/composer.ts
```diff
@@ -1409,692 +1368,634 @@ export class TransactionComposer {
...
+        this.txns.some((t) => isAppCall(t))
+          ? await this.analyzeGroupRequirements(builtTransactions.transactions, suggestedParams, this.composerConfig)
+          : undefined
 
-    if (params.lease) txnParams.lease = encodeLease(params.lease)! satisfies Transaction['lease']
-    if (params.rekeyTo) txnParams.rekeyTo = params.rekeyTo.toString() satisfies Transaction['rekeyTo']
-    const encoder = new TextEncoder()
-    if (params.note)
-      txnParams.note = (typeof params.note === 'string' ? encoder.encode(params.note) : params.note) satisfies Transaction['note']
+      await this.populateResourcesAndGroupTransactions(builtTransactions.transactions, groupAnalysis)
```
neilcampbell:
```suggestion
      await this.populateTransactionAndGroupResources(builtTransactions.transactions, groupAnalysis)
```

src/types/composer.ts
```diff
@@ -1409,692 +1368,634 @@ export class TransactionComposer {
...
-        context: {},
-      }
-    })
-
-    return txnWithSigners
-  }
+  public async build() {
+    if (!this.transactionsWithSigners) {
+      const suggestedParams = await this.getSuggestedParams()
+      const builtTransactions = await this.buildTransactionsSuggestedParamsProvided(suggestedParams)
```
neilcampbell:
```suggestion
      const builtTransactions = await this._buildTransactions(suggestedParams)
```

If you're just trying to have a different name.

src/transaction/legacy-bridge.ts
```diff
@@ -62,21 +62,21 @@ export async function legacySendTransactionBridge<T extends CommonTransactionPar
...
           signer:
             'signers' in transaction ? (transaction.signers.get(i) ?? getSenderTransactionSigner(from)) : getSenderTransactionSigner(from),
         }))
-        .forEach((t) => sendParams.atc!.addTransaction(t))
-      // Populate ATC with method calls
+        .forEach((t) => sendParams.transactionComposer!.addTransaction(t.txn, t.signer))
+      // Populate the composer with method calls
       if ('transactions' in transaction) {
-        transaction.methodCalls.forEach((m, i) => sendParams.atc!['methodCalls'].set(i + baseIndex, m))
+        transaction.methodCalls.forEach((m, i) => sendParams.transactionComposer!['methodCalls'].set(i + baseIndex, m))
```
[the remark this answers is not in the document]
PatrickDinh: yeh it's pretty hacky. The problem comes from here
```
const transaction = await txn(transactionCreator)(params)
```
this causes the `trasaction` to be a `txnWithSigner` => when it's added to the composer, the `methodCalls` map isn't updated.

src/transaction/perform-transaction-composer-simulate.ts
PatrickDinh: When I think about it further, an edge case could be when a user can call `populateAppCallResources` on an `atc` then passes it into this method.

We could add the the migration guide that this method doesn't handle resource population.

src/transaction/transaction.ts
```diff
@@ -425,547 +278,117 @@ async function getGroupExecutionInfo(
...
+    } else {
```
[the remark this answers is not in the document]
PatrickDinh: This is a good call. I refactored the composer to hold transactions with max fees and add a method to set max fees. This is a lot cleaner.

src/types/composer.ts
```diff
@@ -1409,692 +1368,634 @@ export class TransactionComposer {
...
+      } else if (ctxn.type === 'asyncTxn') {
+        transactions.push(await ctxn.data.txn)
+        // Use the signer that was set when the asyncTxn was added
+        if (ctxn.data.signer) {
+          signers.set(transactionIndex, ctxn.data.signer)
+        }
+        transactionIndex++
+      } else {
+        let transaction: Transaction
+        const header = buildTransactionHeader(ctxn.data, suggestedParams, defaultValidityWindow)
```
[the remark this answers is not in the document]
PatrickDinh: The `CommonTransactionParams` has fields such as `maxFee`, `staticFee`. It's slightly different to the `header` which only has the `fee` field. Maybe a better name for `header` is `commonData`?

src/types/composer.ts
```diff
@@ -1409,692 +1368,634 @@ export class TransactionComposer {
...
 
-    return { txn, context: { maxFee: logicalMaxFee } }
+    return { transactions, methodCalls: this.methodCalls, signers }
   }
 
   /**
-   * Builds an ABI method call transaction and any other associated transactions represented in the ABI args.
-   * @param includeSigner Whether to include the actual signer for the transactions.
-   *  If you are just building transactions without signers yet then set this to `false`.
+   * Compose all of the transactions without signers and return the transaction objects directly along with any ABI method calls.
```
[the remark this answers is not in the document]
PatrickDinh: The jsdocs confused me as first too because this method returns the `signer` Map. I think it means that the transactions returned doesn't have signers attached. Should I change the docs?

src/types/composer.ts
```diff
@@ -2116,28 +2017,37 @@ export class TransactionComposer {
...
    */
-  async simulate(options: RawSimulateOptions): Promise<SendAtomicTransactionComposerResults & { simulateResponse: SimulateTransaction }>
-  async simulate(options?: SimulateOptions): Promise<SendAtomicTransactionComposerResults & { simulateResponse: SimulateTransaction }> {
+  async simulate(options: RawSimulateOptions): Promise<SendTransactionComposerResults & { simulateResponse: SimulateTransaction }>
+  async simulate(options?: SimulateOptions): Promise<SendTransactionComposerResults & { simulateResponse: SimulateTransaction }> {
     const { skipSignatures = false, ...rawOptions } = options ?? {}
-    const atc = skipSignatures ? new AtomicTransactionComposer() : this.atc
 
+    const builtTransactions = await this.buildTransactions()
+    const transactions =
```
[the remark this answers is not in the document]
PatrickDinh: No, `buildTransactions` doesn't do grouping. Grouping is applied in `build`, after resource population and fee coverage.

You mentioned this makes me think that the names `buildTransactions` and `build` are confusing. I kept them because:
- `buildTransactions` existed in the composer. It does similar functionality.
- `build` existed in the composer, but private. In `main`, resource population + fee coverage happen in `send`, for this version, I lifted them to `build`
I'm happy to find better names for them but that means breaking changes.

src/types/composer.ts
```diff
@@ -2194,4 +2109,124 @@ export class TransactionComposer {
...
+    const simulateResult = await this.algod.simulateTransaction(simulateRequest)
+    return simulateResult
+  }
+}
+
+function isAppCall(ctxn: Txn): boolean {
+  return (
+    ctxn.type === 'appCall' ||
+    ctxn.type === 'methodCall' ||
+    (ctxn.type === 'txnWithSigner' && ctxn.data.txn.type === TransactionType.AppCall)
```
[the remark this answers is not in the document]
PatrickDinh: I'll remove this function, it isn't needed anymore. In `build` (where it was used) I have access to the built transaction, I can check for the transaction type directly.

src/types/composer.ts
```diff
@@ -547,19 +234,19 @@ export interface BuiltTransactions {
   signers: Map<number, algosdk.TransactionSigner>
 }
 
+class BuildComposerTransactionsError extends Error {
```
PatrickDinh: @neilcampbell I created this error type for errors during build transactions. I think this is helpful because before this, in debug mode, if something goes wrong during resource population, we will need to run simulate again in the error handler. Using this error type, we can surface to simulation result to the error handler => eliminate the need to simulate again.

src/types/composer.ts
```diff
@@ -2152,33 +2075,38 @@ export class TransactionComposer {
             },
           }
         : undefined),
-    } satisfies SimulateRequest)
+    } satisfies SimulateRequest
 
-    const failedGroup = simulateResponse?.txnGroups[0]
-    if (failedGroup?.failureMessage) {
-      const errorMessage = `Transaction failed at transaction(s) ${failedGroup.failedAt?.join(', ') || 'unknown'} in the group. ${failedGroup.failureMessage}`
+    const simulateResponse = await this.algod.simulateTransaction(simulateRequest)
```
PatrickDinh: @neilcampbell The name of the return type of `simulateTransaction` is `SimulateTransaction` while algosdk uses `SimulateResponse`. I think we should update our algod to use `SimulateResponse` too.

src/transaction/transaction.spec.ts
```diff
@@ -1132,7 +1133,14 @@ describe('Resource population: meta', () => {
 
     const result = await externalClient.send.call({
       method: 'createBoxInNewApp',
-      args: [algorand.createTransaction.payment({ sender: testAccount, receiver: externalClient.appAddress, amount: (1).algo() })],
+      args: [
+        algorand.createTransaction.payment({
+          sender: testAccount,
+          receiver: externalClient.appAddress,
+          amount: (1).algo(),
+          signer: testAccount,
```
neilcampbell: This shouldn't be needed, as `fixture.newScope()` should have added the signer account.
Does this indicate some kind of behavioural change?

src/transactions/payment.ts
```diff
@@ -0,0 +1,28 @@
...
+export type PaymentParams = CommonTransactionParams & {
+  /** The address of the account that will receive the Algo */
+  receiver: string | Address
+  /** Amount to send */
+  amount: AlgoAmount
+  /** If given, close the sender account and send the remaining balance to this address
+   *
+   * *Warning:* Be careful with this parameter as it can lead to loss of funds if not used correctly.
+   */
+  closeRemainderTo?: string | Address
```
neilcampbell: This field doesn't appear to be mapped?

Can you double check the mappings for all transaction types?

src/types/composer.spec.ts
```diff
@@ -63,5 +65,97 @@ describe('TransactionComposer', () => {
...
+        assetId: 1337n,
+        sender,
+        receiver: sender,
+      })
+
+      errorTransformers.forEach((errorTransformer) => {
+        composer.registerErrorTransformer(errorTransformer)
+      })
+
+      const simulateResult = await composer.simulate({ throwOnFailure: false })
```
neilcampbell: Instead of `throwOnFailure` with a default of true, would `resultOnFailure` with a default of false be better?

src/types/composer.ts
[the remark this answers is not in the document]
neilcampbell: I see, I think header is fine.

src/types/composer.ts
```diff
@@ -1409,692 +1336,690 @@ export class TransactionComposer {
...
-   * @returns The newly built atomic transaction composer and the transactions
+   * @returns The newly built transaction composer and the transactions
    * @example
    * ```typescript
    * const { atc, transactions, methodCalls } = await composer.rebuild()
    * ```
    */
   async rebuild() {
-    this.atc = new algosdk.AtomicTransactionComposer()
+    this.transactionsWithSigners = undefined
```
neilcampbell: Do we need to also do this.signedTransactions?

If so, i think it make sense to extract this logic into a private `reset()` method, which can be called by `send()` (in the scenario where resource population/fee coverage settings are different)  as well.

src/types/composer.ts
```diff
@@ -2194,4 +2142,104 @@ export class TransactionComposer {
...
+
+  public setMaxFees(maxFees: Map<number, AlgoAmount>) {
+    maxFees.forEach((_, index) => {
+      if (index > this.txns.length - 1) {
+        throw new Error(`Index ${index} is out of range. The composer only contains ${this.txns.length} transactions`)
+      }
+    })
+
+    maxFees.forEach((maxFee, index) => {
+      this.txns[index].data.maxFee = new AlgoAmount({ microAlgos: maxFee.microAlgos })
```
neilcampbell:
```suggestion
      this.txns[index].data.maxFee = maxFee
```

src/types/composer.ts
neilcampbell: I like the idea, however this also become part of the public interface of the composer and an exception that a user may receive. I'd be inclined to keep as before and run another simulate. It only occurs in debug mode, which should only be enabled in testing scenarios (non prod).

src/types/composer.ts
```diff
@@ -582,6 +266,17 @@ export class TransactionComposer {
...
+  private composerConfig: TransactionComposerConfig
+
+  private transactionsWithSigners?: TransactionWithSigner[]
+
+  private signedTransactions?: SignedTransaction[]
+
+  // Note: This doesn't need to be a private field of this class
+  // It has been done this way so that another process can manipulate this values, i.e. `legacySendTransactionBridge`
+  // Once the legacy bridges are removed, this can be calculated on the fly
+  private methodCalls: Map<number, algosdk.ABIMethod> = new Map()
```
neilcampbell: Having to keep this in sync isn't great. I'm wondering if we could instead calculate on the fly (as you mentioned above), however add a private method which the legacy bridge can call?

src/transaction/transaction.ts
```diff
@@ -1172,15 +545,15 @@ export async function getTransactionParams(params: SuggestedParams | undefined,
...
- * @param atc The atomic transaction composer
+ * Returns the array of transactions currently present in the given `TransactionComposer`
+ * @param atc The transaction composer
  * @returns The array of transactions with signers
  */
-export function getAtomicTransactionComposerTransactions(atc: AtomicTransactionComposer) {
+export async function getTransactionComposerTransactions(composer: TransactionComposer) {
   try {
-    return atc.clone().buildGroup()
+    return (await composer.clone().build()).transactions.map((transactionWithSigner) => transactionWithSigner.txn)
```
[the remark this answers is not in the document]
PatrickDinh: this method was removed

src/transaction/transaction.ts
```diff
@@ -312,547 +278,60 @@ async function getGroupExecutionInfo(
...
  * Parts of this function will eventually be implemented in algod. Namely:
  * - Simulate will return information on how to populate reference arrays, see https://github.com/algorand/go-algorand/pull/6015
  */
 export async function prepareGroupForSending(
-  atc: algosdk.AtomicTransactionComposer,
-  algod: AlgodClient,
+  composer: TransactionComposer,
   sendParams: SendParams,
-  additionalAtcContext?: AdditionalAtomicTransactionComposerContext,
+  additionalAtcContext?: AdditionalTransactionComposerContext,
```
neilcampbell:
```suggestion
  additionalContext?: AdditionalTransactionComposerContext,
```

neilcampbell (approved):

PR #466, opened 2025-11-27
Abi integration

PatrickDinh:
## Proposed Changes

-
-
-


packages/abi/src/abi-method.ts
```diff
@@ -19,49 +19,164 @@ export enum ABIReferenceType {
...
+      /** The raw return value as bytes.
+       *
+       * This value will be empty if the method does not return a value (return type "void")
+       */
+      rawReturnValue: Uint8Array
+      /** The parsed ABI return value.
+       *
+       * This value will be undefined if decoding failed or the method does not return a value (return type "void")
+       */
+      returnValue: ABIValue
```
neilcampbell: Should this be optional?

packages/abi/src/abi-method.ts
```diff
@@ -19,49 +19,164 @@ export enum ABIReferenceType {
...
+        }
+      }
+    }
+
+    if (argsEnd === -1) {
+      throw new Error(`Invalid method signature: ${signature}`)
+    }
+
+    const name = signature.slice(0, argsStart)
+    const args = parseTupleContent(signature.slice(argsStart + 1, argsEnd)) // hmmm the error is bad
```
neilcampbell: > // hmmm the error is bad

?

src/types/app-spec.ts
```diff
@@ -161,6 +173,43 @@ export interface AppSpec {
   bare_call_config: CallConfig
 }
 
+interface ABIContractParams {
```
neilcampbell: Not sure ABI is the correct term to use here?
ABI mainly refers to the encoding standard. Maybe AppSpecParams? Same applies for ABIContractNetworks and ABIContractNetworkInfo, however the rest of the names seem ok to me.

Potentially this also allows us to drop the Param suffix?

packages/abi/src/abi-method.ts
PatrickDinh: This comment was copied over from `algokit-core`. I'm not sure the context now to be honest. I had a look at the error messages inside `parseTupleContent` and they look ok to me. 

One thing jumped out is that in `abi-type.ts` we use prefix for error messages, for example `Validation Error: the content should not have consecutive commas`. The rest of `utils-ts` doesn't do that.

src/types/app-spec.ts
PatrickDinh: The names were copied from `algosdk` over. I think `ContractDefinition` is a better name. What do you think?

packages/abi/src/abi-method.ts
```diff
@@ -19,49 +19,164 @@ export enum ABIReferenceType {
...
 }
 
+/** Represents an ABI method return value with parsed data. */
+export type ABIReturn =
+  | {
+      /** The method that was called. */
+      method: ABIMethod
+      /** The raw return value as bytes.
+       *
+       * This value will be empty if the method does not return a value (return type "void")
```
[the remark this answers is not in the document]
PatrickDinh: I missed the comments in this type when integrating ABI features into utils. I think `empty` and `undefined` aren't applicable anymore.

- The method returns `undefined` when ABIMethod returns `void`. It doesn't return `ABIReturn`.
- There is another union variant that describes the decoding error case.

I updated the comments to reflex this fact.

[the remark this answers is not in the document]
PatrickDinh: It's a previous behaviour
https://github.com/algorandfoundation/algokit-utils-ts/blob/3d899a4e5538b3899d738f1db32274d027930292/src/types/app-manager.ts#L432
and the type for `ABIReturn` is 
```
/** The return value of an ABI method call */
export type ABIReturn =
  | {
      rawReturnValue: Uint8Array
      returnValue: ABIValue
      method: ABIMethod
      decodeError: undefined
    }
  | { rawReturnValue?: undefined; returnValue?: undefined; method?: undefined; decodeError: Error }

```

packages/abi/src/abi-type.ts
```diff
@@ -1,718 +1,904 @@
...
+   */
+  static from(str: string): ABIType {
+    if (str.endsWith('[]')) {
+      const childType = ABIType.from(str.slice(0, str.length - 2))
+      return new ABIArrayDynamicType(childType)
+    }
+    if (str.endsWith(']')) {
+      const stringMatches = str.match(STATIC_ARRAY_REGEX)
+      if (!stringMatches || stringMatches.length !== 3) {
+        throw new Error(`malformed static array string: ${str}`)
```
neilcampbell:
```suggestion
        throw new Error(`Malformed static array string: ${str}`)
```

We tend to start error message with uppercase. This applies to all the changes removing the prefix.

neilcampbell (approved):

PR #481, opened 2025-12-09
fix: adjust resource population for new account refs max

neilcampbell:


src/transactions/app-call.ts
```diff
@@ -538,10 +533,10 @@ function populateGroupResource(
 
     switch (resource.type) {
       case GroupResourceType.Account:
-        return accountsCount < MAX_ACCOUNT_REFERENCES
+        return accountsCount + assetsCount + appsCount + boxesCount < MAX_OVERALL_REFERENCES
```
PatrickDinh: should this be `accountsCount < MAX_ACCOUNT_REFERENCES & accountsCount + assetsCount + appsCount + boxesCount < MAX_OVERALL_REFERENCES`?
Additionally, should we update the constant `MAX_APP_CALL_ACCOUNT_REFERENCES` to `8` to match with main?

PatrickDinh (approved):

PR #522, opened 2026-01-16
Docs/add addtional typedoc tags

lempira:
commit 1: Add @throws JSDoc tags to document error conditions across ABI, transaction, and mnemonic packages
commit 2: Add @remarks JSDoc tags to provide implementation context for key classes (MultisigAccount, LogicSigAccount, ABIType, AppDeployer)
commit 3: Add @see JSDoc tags to create cross-references between related functions and classes


p2arthur (approved):
I think these new tags bring good value. Worth adding to the other repos as well IMO

Issue #313, opened 2024-09-06
Time management utilities


cusma:
## Problem

For some use cases, particularly when Smart Contract are based on time or blocks' height conditions, it is useful to conveniently "wait" for a given timestamp or block.

Moreover, when thing run on a LocalNet in DevMode it is possible to "fast forward" both time and blocks.

It would be very useful if such methods could have a nice UX (a less boilerplate code). 

## Solution

Add to `AlgorandClient` and `LocalNet` (or some better place) convenient methods to fetch time/blocks, wait for time/blocks conditions, jump to time/block (in DevMode).

### Proposal

Add to the `AlgorandClient` the following methods:

- `getLastRound`: returns the last committed round from `Algod`
- `getLatestTimestamp`: returns the latest UNIX timestamp committed with the last round from `Algod`
- `waitUntilRound`: is a "sleep" conditions that alts the execution until a given round is committed (it could leverage `getLastRound`)
- `waitUntilTimestamp`: is a "sleep" conditions that alts the execution until a given UNIX timestamp is committed (it could leverage `getLatestTimestamp`)

In order to speed up tests when things run LocalNet (in DevMode), the following methods can be added (not in `AlgorandClient`, since it's DevMode specific):

- `blockWarp`: generate blocks to jump directly to a given blocks' height (blocks are instantly generated as soon as a transaction is committed in DevMode)
- `timeWarp`: set a proper time offset to commit a block with the desired UNIX timestamp.

Specifically for `timeWarp` I'd suggest the following implementation to jump directly to `toTimestamp`:

1. Get the `latestTimestamp` with `getLatestTimestamp`
3. Set the timestamp offset with the `Algod` method to: `toTimestamp` - `latestTimestamp`
4. Generate a block to commit the desired offset (this could use `blockWarp` with a single block generation)
5. Reset the timestamp offset to `0`

### Pros and Cons

**Pros:** speed up test, less boilerplate code, nicer UX.
**Cons:** none.

## Dependencies

Algorand SDK and LocalNet in DevMode

robdmoore:
My suggestion for the design of this would be a new `NetworkManager` class is added (similar to `AppManager`, `AssetManager`, etc.) with the following signatures from `AlgorandClient`:

* `algorand.network.getLastRound()`
* `algorand.network.getLatestTimestamp()`
* `algorand.network.waitUntilRound()`
* `algorand.network.waitUntilTimestamp()`
[...]

PR #534, opened 2026-02-02
Feature - time management

PatrickDinh:
## Proposed Changes

- Add time management features

This should address #313


src/network-manager.ts
```diff
@@ -0,0 +1,225 @@
...
+   */
+  async waitUntilTimestamp(targetTimestamp: bigint): Promise<void> {
+    // Note: this method reduces the number of calls to algod
+    //  by estimating the number of rounds to wait first
+    //  then wait for the target round to be committed
+    //  then poll every second for the right timestamp
+
+    // BLOCK_TIME_SECONDS to be slightly lower than the average block time so that it will undershoot
+    const BLOCK_TIME_SECONDS = 2.7
+    const POLLING_INTERVAL_MS = 1000
```
joe-p: Maybe these could be optional params?

src/network-manager.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    while (true) {
```
joe-p: Maybe we should add a timeout here just to be safe?

joe-p (commented):
Some small comments, but overall LGTM. Before merging I want to wait to get feedback from the team tomorrow so we can plan the work for python as well.

src/network-manager.ts
PatrickDinh: good idea, I added 5 seconds timeout as default. I have changed the logic a bit so that the timeout only affects the polling bit, not the wait for block bit. The user can supply their own value.

joe-p (approved):

PR #536, opened 2026-02-03
Refactor: centralize example runner and auto-build dist

mrcointreau:
## Summary                                                                                                                                                      
                                                                                                                                                               
  - Add `run-example.mjs` script that resolves example files by path or bare filename, auto-builds `dist/` if missing, and runs them via `tsx`                       
  - Replace inline `npx tsx --tsconfig ../tsconfig.run.json` calls in all `verify-all.sh` scripts with `npm run example --`                                          
  - Move `tsx` from implicit `npx` usage to an explicit `devDependency`                                                                                              
  - Remove the separate "Build package" CI step since the example runner handles it                                                                            
                                                                                                                                                               
  ## Test plan                                                                                                                                                    
                                                                                                                                                               
  - Run `npm run example -- transact/01-payment-transaction.ts` from the examples dir                                                                            
  - Run a `verify-all.sh` script to confirm it still works end-to-end                                                                                            
  - Delete` dist/` and verify the auto-build triggers

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -60,9 +60,6 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Build package
-        run: npm run build
-
       - name: Start LocalNet
         run: |
           pipx install algokit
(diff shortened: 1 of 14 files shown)


lempira (approved):
lgtm

PR #537, opened 2026-02-04
Feat/updated docs with starlight

lempira:
The documentation system has been completely restructured using Starlight as the framework, with starlight-typedoc handling API documentation generation from multiple entry points (main, testing, abi, transact, etc.). Concept documentation is now organized into logical subfolders (core/, building/, advanced/) with an enhanced homepage featuring installation instructions and a quick-start example.

The standalone TypeDoc configuration and generated API docs have been removed from the repository. API documentation is now generated automatically as part of the Starlight build process, meaning it no longer needs to be committed separately - the npm run docs:build command handles both the curated content and API reference generation in one step.

A temporary docs-preview.yml workflow has been added to publish the Starlight docs from this branch for review. This workflow should be removed after merge, and release.yml should be updated to publish from main.

https://algorandfoundation.github.io/algokit-utils-ts/

image


docs/.vscode/launch.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "command": "./node_modules/.bin/astro dev",
+      "name": "Development server",
```
PatrickDinh: I guess this is related to the docs page? Maybe we should use a name more related to docs, for example "Start docs server" (just a suggestion)

PatrickDinh (approved):

docs/.vscode/launch.json
lempira: Updated to `Utils-ts Docs server`

PR #547, opened 2026-02-24
feat!: wrapped secrets

joe-p:
This PR adds new interfaces for wrapping and unwrapping ed25519 secrets. The following secrets are supported and have their own interfaces:

* 32-byte seed
* 64-byte `scalar || prefix`
* 96-byte extended private key (`scalar || prefix || chain_code`)

A function was added for taking a union of these interfaces and generating a signer, which can then be used to generate an address and algorand-specific signing functions. 

The main idea is to be compatible with a secrets manager without having to keep the secret loaded in memory throughout the entire lifetime of the signing function. 

The tests demonstrates how it is used:

```ts
  test('addr with signers from wrapped seed', async () => {
    const seed = ed.utils.randomSecretKey()
    const wrappedSeed = {
      unwrapEd25519Seed: async () => new Uint8Array(seed),
      wrapEd25519Seed: async () => {},
    }
    const signingKey = await nobleEd25519SigningKeyFromWrappedSecret(wrappedSeed)
    const addressWithSigners = generateAddressWithSigners(signingKey)

    await runTests(addressWithSigners, signingKey.ed25519Pubkey)
  })

  test('addr with signers from wrapped scalar || prefix', async () => {
    const { accountGenerator } = await peikertXHdWalletGenerator()
    const generated = await accountGenerator(0, 0)
    const wrappedScalarAndPrefix = {
      unwrapHdScalarAndPrefix: async () => new Uint8Array(generated.extendedPrivateKey.slice(0, 64)),
      wrapHdScalarAndPrefix: async () => {},
    }
    const signingKey = await nobleEd25519SigningKeyFromWrappedSecret(wrappedScalarAndPrefix)
    const addressWithSigners = generateAddressWithSigners(signingKey)

    await runTests(addressWithSigners, signingKey.ed25519Pubkey)
  })

  test('addr with signers from wrapped extended private key', async () => {
    const { accountGenerator } = await peikertXHdWalletGenerator()
    const generated = await accountGenerator(0, 0)
    const wrappedExtendedPrivateKey = {
      unwrapHdExtendedPrivateKey: async () => new Uint8Array(generated.extendedPrivateKey),
      wrapHdExtendedPrivateKey: async () => {},
    }
    const signingKey = await nobleEd25519SigningKeyFromWrappedSecret(wrappedExtendedPrivateKey)
    const addressWithSigners = generateAddressWithSigners(signingKey)

    await runTests(addressWithSigners, signingKey.ed25519Pubkey)
  })
```

More comprehensive documentation and best practices will be included in the key management documentation that is currently being worked on


ehanoc:
I'm wondering if we should simply support: 

`
64-byte scalar (LH) || RH (proper term)
`
`
96-byte extended private key (scalar (KL) || RH (or known as KR) || chain_code)
` 

Including the `seed` is a convenience that can be missed used by the dev; once derived keys should be ideally handled individually.  

I'm not against including it, as long as we make the case that the other forms are preferable and the responsibility of on the dev to handle and derive keys from the seeds. Since they already have xHD libs for that anyway... we don't need to own that.

joe-p:
Also we could just support 96 byte extended key. If someone wants to store their key without the chaincode they can just pad with 32 bytes during the unwrap.

ehanoc:
> Also we could just support 96 byte extended key. If someone wants to store their key without the chaincode they can just pad with 32 bytes during the unwrap.

i like it. Lets do that

ehanoc:
> > 64-byte scalar (LH) || RH (proper term)
> 
> Maybe we can call it `hashedSeed` instead of `scalarAndPrefix`?

We could if document it really well. And explain it there that is composed LH || RH ; and that LH is a scalar that could the result of BIP32 derivation, etc ,etc.

ehanoc (approved):

ehanoc:
looks good @joe-p . I think there's a conflict, so once its solved, we can merge

PR #550, opened 2026-03-02
docs: signing examples & secret management docs

joe-p:
Adds 3 examples for working with ed25518 seed, xHD extended private key, and AWS KMS. 

Also includes documentation for secret management which refers to these examples. 

Based on conversations with Marco, we probably want to add a utiltiy function that does the wrap/unwrap, generate ed25519 signing key, generate algorand address with signers in one call but that will be in a follow-up PR


mrcointreau (approved):
Pushed a couple of commits for:
- adding signing in the sidebar config and signig examples to the astro docs                                                                     
- updating the `README.md` with the new signing examples                                                                           
- adding crypto package path mapping to `examples/tsconfig.json`                                                                                             
- including signing in the top-level `verify-all.sh` suite list
- adding eslint -> flatted vuln to `.nsprc`

LGTM

PR #575, opened 2026-05-12
feat!: AccountManager.fromSecret with simplified wrapped interfaces

joe-p:
* simplifies the wrapped interfaces and makes wrap function optional
* add wrapped mnemonic interfaces
* Moves algo25 under crypto to avoid circular deps

Open question: should we just remove algo25 package all together and just export the functionality from crypto?


packages/crypto/src/index.ts
```diff
@@ -1,16 +1,17 @@
...
 import { bytesToNumberLE, numberToBytesLE } from '@noble/curves/utils.js'
 import { mod } from '@noble/curves/abstract/modular.js'
+import { seedFromMnemonic } from './algo25/'
 
 export * from './ed25519'
 export * from './hash'
 export * from './hd'
 
-export type WrappedEd25519Secret = WrappedEd25519Seed | WrappedHdExtendedPrivateKey
+export type WrappedEd25519Secret = WrappedEd25519Seed | WrappedHdExtendedPrivateKey | WrappedHdMnemonic | WrappedLegacyMnemonic
```
iglosiggio, line 14: Unrelated question: why do we prefer shape-differentiated variants instead of variants with a common shape and a discriminator key?

So basically:
```ts
export type WrappedEd25519Secret = WrappedEd25519Seed | WrappedHdExtendedPrivateKey | WrappedHdMnemonic | WrappedLegacyMnemonic
```
but we change each variant to:
```ts
export type WrappedHdMnemonic = {
  kind: 'hd'
  unwrap: () => Promise<string>
  wrap?: () => Promise<void>
}
export type WrappedHdExtendedPrivateKey = {
  kind: 'hd-extended-pk'
  unwrap: () => Promise<Uint8Array>
  wrap?: () => Promise<void>
}
export type WrappedLegacyMnemonic = {
  kind: 'legacy'
  unwrap: () => Promise<string>
  wrap?: () => Promise<void>
}
// ...
```

This alternative shape allows for common code to handle the variants (and can help exhaustiveness checks if you go the switch route instead of the if/elseif route when writing code).

iglosiggio (approved):
Changes look ok. I personally feel that `WrappedEd25519Secret` should be an internal API. The current design is cumbersome to consume and most changes to it will break the code that consumes it (adding a variant, changing the return type of an unwrap function, etc).

packages/crypto/src/index.ts
joe-p: The main reason is that if a caller of the function forgets to check `kind` they could potentially misuse the secret which will definitely lead to unexpected behavior which could potentially lead to a loss of funds or security issue. The friction is intentional here. 

That being said if we want to add a `kind` property and use that instead of `in` checks that is fine with me.


I added an exhaustiveness check in [6056462](https://github.com/algorandfoundation/algokit-utils-ts/pull/575/commits/6056462cd2c34c32b652e8adda3b0636193b10c0)

PR #576, opened 2026-05-13
fix: properly match against new algod errors

joe-p:
algod 4.7 changed the fee error from this:

```go
return fmt.Errorf("fee too small %#v", cx.subtxns)
```

To this:

```go
return fmt.Errorf("group fee %s too small (need %s) %#v", groupPaid, groupFee, cx.subtxns)
```

Matching on `too small` allows us to capture both

### DIFF
--- a/src/app-client.ts
+++ b/src/app-client.ts
@@ -1397,7 +1397,7 @@ export class AppClient {
             const error = e as Error
             // For read-only calls with max opcode budget, fee issues should be rare
             // but we can still provide helpful error message if they occur
-            if (params.coverAppCallInnerTransactionFees && error && error.message && error.message.match(/fee too small/)) {
+            if (params.coverAppCallInnerTransactionFees && error && error.message && error.message.match(/too small/)) {
               throw Error(`Fees were too small. You may need to increase the transaction maxFee.`)
             }
             throw e
--- a/src/composer.ts
+++ b/src/composer.ts
@@ -1705,7 +1705,7 @@ export class TransactionComposer {
 
     // Handle any simulation failures
     if (groupResponse.failureMessage) {
-      if (analysisParams.coverAppCallInnerTransactionFees && groupResponse.failureMessage.includes('fee too small')) {
+      if (analysisParams.coverAppCallInnerTransactionFees && groupResponse.failureMessage.includes('too small')) {
         throw new Error(
           'Fees were too small to resolve execution info via simulate. You may need to increase an app call transaction maxFee.',
         )
--- a/src/transaction/transaction.spec.ts
+++ b/src/transaction/transaction.spec.ts
@@ -152,7 +152,7 @@ describe('transaction', () => {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } catch (e: any) {
       const messageRegex = new RegExp(
-        `transaction ${txn2.txId()}: overspend \\(account ${testAccount}, data \\{.*\\}, tried to spend \\{9999999999999\\}\\)`,
+        `transaction ${txn2.txId()}: overspend \\(account ${testAccount}, data \\{.*\\}, tried to spend \\{?9999999`,
       )
       expect(e.traces[0].message).toMatch(messageRegex)
     }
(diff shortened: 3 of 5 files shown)


src/composer.ts
```diff
@@ -1705,7 +1705,7 @@ export class TransactionComposer {
 
     // Handle any simulation failures
     if (groupResponse.failureMessage) {
-      if (analysisParams.coverAppCallInnerTransactionFees && groupResponse.failureMessage.includes('fee too small')) {
+      if (analysisParams.coverAppCallInnerTransactionFees && groupResponse.failureMessage.includes('too small')) {
```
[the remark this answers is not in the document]
joe-p: It's the only too small error the AVM will throw

src/transaction/transaction.spec.ts
```diff
@@ -152,7 +152,7 @@ describe('transaction', () => {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } catch (e: any) {
       const messageRegex = new RegExp(
-        `transaction ${txn2.txId()}: overspend \\(account ${testAccount}, data \\{.*\\}, tried to spend \\{9999999999999\\}\\)`,
+        `transaction ${txn2.txId()}: overspend \\(account ${testAccount}, data \\{.*\\}, tried to spend \\{?9999999`,
```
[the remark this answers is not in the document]
joe-p: This is because this error also changed

src/app-client.ts
```diff
@@ -1397,7 +1397,7 @@ export class AppClient {
             const error = e as Error
             // For read-only calls with max opcode budget, fee issues should be rare
             // but we can still provide helpful error message if they occur
-            if (params.coverAppCallInnerTransactionFees && error && error.message && error.message.match(/fee too small/)) {
+            if (params.coverAppCallInnerTransactionFees && error && error.message && error.message.match(/too small/)) {
```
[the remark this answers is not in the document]
joe-p: It's the only too small error the AVM will throw

src/transaction/transaction.spec.ts
joe-p, line 155: Just noting that this error also changed, but it only affects this one test

SilentRhetoric: It appears this new regex has an incomplete group structure that is missing a closing `)`.

Should it be as follows?

`transaction ${txn2.txId()}: overspend \\(account ${testAccount}, data \\{.*\\}, tried to spend \\{?9999999\\)`

SilentRhetoric (dismissed):

src/transaction/transaction.spec.ts
joe-p: The problem is that the new error message includes a decimal point. The two options were either: A) Update the regex to handle either version or B) truncate what we match. 

Since this is just a test and we aren't actually using the match, it felt easier to just truncate

iglosiggio: In any case: why the optional `{`? it feels like you want either:
```js
`... tried to spend \\{9999999`,
```
or
```js
`tried to spend 9999999`,
```

iglosiggio (approved):

src/transaction/transaction.spec.ts
joe-p: The error uses `%v` in Go's Sprintf to print `basics.MicroAlgo`. The new version of go-algorand implements a String function. This means that the older error includes a literal `{` from the struct printing but the new error doesn't

```go
func (a MicroAlgos) String() string {
	if a.Raw == 0 {
		return "0.0A"
	}
	if a.Raw < 1_000 {
		return fmt.Sprintf("%duA", a.Raw)
	}
	if a.Raw < 1_000_000 {
		millis := a.Raw / 1_000
		fraction := a.Raw % 1_000
		if fraction == 0 {
			return fmt.Sprintf("%dmA", millis)
		}
		return fmt.Sprintf("%d.%03dmA", millis, fraction)
	}
	whole := a.Raw / 1_000_000
	fraction := a.Raw % 1_000_000
	if fraction == 0 {
		return fmt.Sprintf("%dA", whole)
	}
	return fmt.Sprintf("%d.%06dA", whole, fraction)
}
```

joe-p: I suppose since this is just test code we could require the latest algod, but in library code we need to support any algod version that has the latest conensus.

PR #586, opened 2026-07-20
fix(transact): support empty access references

kylebeee:
## Summary

- encode empty access-list resource references using the canonical empty map representation
- preserve canonical empty references when decoding transactions
- decode the previous non-canonical empty box representation for compatibility
- add focused wire-level and round-trip regression tests

## Root cause

go-algorand permits an empty ResourceRef in txn.Access to request additional box I/O budget. The transact codec represented that resource in memory as a box with app ID zero and an empty name, but encoded it as a nested empty BoxRef. The same codec then rejected the representation because the omitted box name was treated as missing.

The nested empty BoxRef and canonical empty ResourceRef produce different MessagePack bytes. Signing the non-canonical bytes can therefore produce a payload that does not match the transaction encoding used by algod for signature verification.

## Impact

Empty access references now encode canonically, survive decode/encode round trips, and no longer fail with "Access list box reference is missing name". Existing callers can continue using the current in-memory empty-box sentinel.

## Validation

- focused empty access-reference tests: 2 passed
- transact TypeScript check
- transact ESLint check
- Prettier check
- full repository build
- independent parity check against algosdk 3.6.0: identical 153-byte transaction encoding and identical transaction ID

### DIFF
--- a/packages/transact/src/transactions/transaction-meta.ts
+++ b/packages/transact/src/transactions/transaction-meta.ts
@@ -337,12 +337,14 @@ class AppCallDataCodec extends Codec<AppCallTransactionFields | undefined, Recor
         if (box.appId && box.appId !== appId) {
           appIndex = ensure({ appId: box.appId })
         }
-        accessList.push({
-          b: {
-            i: numberCodec.encodeOptional(appIndex, format),
-            n: bytesCodec.encodeOptional(box.name, format),
-          },
-        })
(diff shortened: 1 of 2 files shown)


kylebeee:
Superseded by #587, which contains the same codec fix on the correctly named branch `fix/empty-access-references` plus the polytest CI version alignment.

PR #587, opened 2026-07-20
fix(transact): support empty access references

kylebeee:
## Summary

- encode empty access-list resource references using the canonical empty map representation
- preserve canonical empty references when decoding transactions
- decode the previous non-canonical empty box representation for compatibility
- add focused wire-level and round-trip regression tests
- align CI's polytest binary with the 0.8 test configurations consumed from algokit-polytest

## Root cause

go-algorand explicitly permits an empty `ResourceRef` in `txn.Access` as the quota-only representation for an additional box I/O budget unit. Duplicate named box references can provide equivalent quota, but empty references express that no additional resource is being named.

The transact codec represents an empty resource in memory as a box with app ID zero and an empty name. It encoded that sentinel as a nested empty `BoxRef`, while its decoder required every nested box reference to contain a name. Consequently, it could neither decode the canonical empty-map representation nor round-trip its previous representation.

Separately, the shared Node CI workflow still defaulted to polytest 0.7.0, while every configuration on `algokit-polytest#main` now declares version 0.8.0.

## Impact

Empty access references now encode canonically, survive decode/encode round trips, and no longer fail with `Access list box reference is missing name`. Existing callers can continue using the current in-memory empty-box sentinel.

## Validation

- focused empty access-reference tests: 2 passed
- transact TypeScript check
- transact ESLint check
- Prettier check
- full repository build
- independent parity check against algosdk 3.6.0: identical 153-byte transaction encoding and identical transaction ID
- `npm run polytest` with polytest 0.8.0

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -17,6 +17,7 @@ jobs:
       test-script: npm run test -- --silent
       install-algokit: true
       polytest: true
+      polytest-version: "0.8.0"
 
   check-docs:
     uses: algorandfoundation/algokit-shared-config/.github/workflows/check-docs.yml@main
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,7 @@ jobs:
       test-script: npm run test -- --silent
       install-algokit: true
       polytest: true
+      polytest-version: "0.8.0"
 
   check_docs:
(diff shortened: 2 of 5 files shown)


joe-p:
We are going to be reverting back to the last stable release (v9) so we are stopping development on the v10 beta. Since v9 uses algosdk for transactions, we will need to ensure that algosdk supports this

PR #600, opened 2026-08-20
chore: refresh npm audit exclusions for current advisories

mitre88:
`node-ci` on current `main` fails in `better-npm-audit` against advisories that landed after #589. Same class of lockfile/CI noise as the GitPython pin in algokit-utils-py#327.

This PR only updates `.nsprc` / `package.json` overrides (and the matching lockfile brace-expansion pin). No runtime source changes.

- Rename stale exclusion `1130705` → current id `1130734` (same GHSA).
- Exclude remaining CI-only / npm-bundled paths (`undici`, `ip-address`, `js-yaml`, `nanoid`) that overrides cannot reach because they live inside npm's bundledDependency tree or are out of range for eslint/postcss.
- Bump `brace-expansion` override to `>=5.0.9`; drop unused `fast-uri` override (not in this lockfile).

Dev-only; nothing here ships in the published package. Review date on the exclusions is 2026-09-01, matching the existing entries.


mrcointreau:
Thanks for this and for the clear write-up on the changes. One process note though: dependency management is handled internally by the AlgoKit team, as a rule we don't accept external PRs of this kind. Please do keep contributing though: bug reports, docs and fixes in `src/` are all very welcome. Thanks again 🙏

PR #607, opened 2026-08-24
feat: support resizing an app as part of an update for AVM 13

tasosbit:
## Proposed Changes

- Add an `appSize` param to `AppUpdateParams` and thread it through the transaction composer, `algorand.createTransaction.appUpdate`/`appUpdateMethodCall`, `algorand.send.*`, `AppClient.params/createTransaction/send.update` (bare and ABI), and the `AppFactory`/`AppDeployer` update params — AVM 13 (consensus v42) makes an app's global state schema and extra program pages mutable, so growing an app no longer requires deleting and recreating it.
- `appSize` is a single optional value holding `globalInts`, `globalByteSlices` and `extraProgramPages`, because the AVM sets all three together and applies them as the app's whole new size rather than as an increase. Leaving it undefined builds the transaction exactly as before and the app keeps its current size.
- Bump the `algosdk` peer dependency to `^3.7.0`, the first version whose `AtomicTransactionComposer` permits these fields on an update call (3.6.0 and earlier reject them outright, which the ABI update path goes through).

### Notes

- The three counts are grouped rather than exposed as separate optional params so that specifying a partial size is a type error instead of a runtime rejection from the AVM (an update carrying only `extraProgramPages` asks the AVM to shrink the global schema to zero and is rejected). A both-or-nothing union of two top-level params was the alternative, but TypeScript collapses a union when it is spread into another object literal, so composing update params with `{...updateParams}` — which `AppDeployer` and most callers do — would have lost the correlation and stopped compiling.
- The local state schema stays immutable on update; the AVM rejects a non-zero local schema on an update outright, so `appSize` deliberately has no local counts.
- `AppDeployer`'s schema-break detection is unchanged: a global state schema increase in the app spec is still treated as a break and handled by `onSchemaBreak`. Teaching `deploy` to grow an existing app in place instead of replacing it is a follow-up, since it changes deployment semantics.

### Testing

- Added LocalNet tests covering a bare update that expands the global state schema, an update that adds extra program pages, an update with no `appSize` leaving the app unchanged, an ABI method call update via `AppClient` and an `AppFactory.deploy` update — all verified against the resulting on-chain app record.
- Full suite passes (405 passed, 2 skipped) against LocalNet running Algorand 5.0 / AVM 13.

### DIFF
--- a/docs/src/content/docs/capabilities/app.md
+++ b/docs/src/content/docs/capabilities/app.md
@@ -143,6 +143,10 @@ The base type for specifying an app update transaction is `AppUpdateParams` (ext
 - `onComplete?: algosdk.OnApplicationComplete.UpdateApplicationOC` - On Complete can either be omitted or set to update
 - `approvalProgram: Uint8Array | string` - The program to execute for all OnCompletes other than ClearState as raw teal that will be compiled (string) or compiled teal (encoded as a byte array (Uint8Array)).
 - `clearStateProgram: Uint8Array | string` - The program to execute for ClearState OnComplete as raw teal that will be compiled (string) or compiled teal (encoded as a byte array (Uint8Array)).
+- `schema?` - The global state schema to resize the app to. It is an object with:
+  - `globalInts: number` - The number of integers saved in global state.
+  - `globalByteSlices: number` - The number of byte slices saved in global state.
+- `extraProgramPages?: number` - The number of extra pages to allocate for the programs.
 
 If you pass in `approvalProgram` or `clearStateProgram` as a string then it will automatically be compiled using Algod and the compilation result will be available via `algorand.app.getCompilationResult` (including the source map). To skip this behaviour you can pass in the compiled TEAL as `Uint8Array`.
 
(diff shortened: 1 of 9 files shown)


tasosbit:
was not meant to go out :)

docs/src/content/docs/capabilities/app.md
```diff
@@ -194,6 +198,35 @@ await algorand.send.appUpdateMethodCall({
 })
 ```
 
+#### Resizing an app
+
+`schema` and `extraProgramPages` require AVM 13 (consensus v42) or later; before that an app's size is fixed when it's created. Both can only be increased: the global state schema can't be reduced below the state the app has already stored, the app can't be given fewer pages than its programs need, and the local state schema stays immutable. Growing an app also costs minimum balance, since the account sending the update takes on the app's size sponsor role and needs to hold the minimum balance for its extra program pages and global state slots.
+
+If neither property is specified the app keeps its current size. The AVM applies both properties together, so if you specify only one of them then the other is automatically resolved to the app's current on-chain value (which costs an extra algod call to look the app up).
```
tasosbit, line 205: better to have a both-or-nothing type and skip the automatic discovert

PR #608, opened 2026-08-24
feat: allow schema and extra page changes through deployer

joe-p:
With v42 we can change the global schema and extra pages. This PR adds support for doing so manually. It also updates the deployer to only consider the schema to be broken upon local schema changes


tasosbit:
It looks like to change either global schema or pages, you need to declare them both

So I think we'd be better off with a shape that requires them both or neither of them

---

edit: after chatting about this a bit, it looks like the best option might be to nest both of these under a new prop that makes it **explicit** that the user is resizing the app, something like 

`resizeApp?: { globalByteSlices: number, globalInts: number, extraProgramPages: number }`

this would make it more obvious that the user is resizing the application, which has the side-effect of making them the size sponsor (taking on the app mbr.)

PR #611, opened 2026-09-02
fix(v10): resolve transaction IDs from the signed transactions

tasosbit:
## Proposed Changes

- Fixes `TransactionComposer.send()` hanging when a signer mutates the transactions it is given. Signers (e.g. wallets) can mutate what they sign, so the built transactions aren't necessarily the ones that reach the network. The mutated group was sent successfully, but `send()` then waited for confirmation of the pre-mutation transaction ID, which is never found.
- Resolves the transactions from the signed transactions instead, so `txIds`, `transactions`, `groupId` and the wait-round calculation all reflect what was actually sent. `simulate()` had the same stale `txIds` problem and is fixed alongside it.
- Keeps the built transaction when its ID is unchanged, as decoding is lossy for zero valued fields — a built `fee` of `0n` decodes back as `undefined`. Without this, the fee assertions in the inner transaction fee coverage tests regress.
- Prefers the signed transactions in the error path too, so debug traces are of what was really sent.
- Adds regression tests with a mutating signer covering a single transaction, a regrouped group of two, and `simulate()`. All three fail on `v10-backup` (the two send cases time out waiting for confirmation) and pass with the fix.

## Notes

This is on the v10 line, which should not affect many downstream consumers. One notable v10 user is Lora, which has an issue when keyregging for a Pera PQ account (common path via nodekit.) The mutated group is submitted but waiting for confirmation fails. 

`main` is not affected in the same way: it delegates to algosdk's `AtomicTransactionComposer`, whose `gatherSignatures()` already derives `txIDs` by decoding the signed transactions. It does have a related stale-ID issue in `sendAtomicTransactionComposer` and `simulate()`, but that is a separate codebase and out of scope here.

### DIFF
--- a/src/composer.spec.ts
+++ b/src/composer.spec.ts
@@ -3,10 +3,12 @@ import { Address } from '@algorandfoundation/algokit-common'
 import {
   decodeSignedTransaction,
   generateAddressWithSigners,
+  groupTransactions,
   Transaction,
   TransactionSigner,
   TransactionType,
 } from '@algorandfoundation/algokit-transact'
+import { Buffer } from 'buffer'
 import { beforeEach, describe, expect, test, vi } from 'vitest'
 import { algo, AlgoAmount } from './amount'
 import { algorandFixture } from './testing'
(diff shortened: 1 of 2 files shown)


joe-p:
Closing in favor of #612

PR #612, opened 2026-09-03
fix: resolve transaction IDs from the signed transactions

tasosbit:
# fix: resolve transaction IDs from the signed transactions

## Problem

Signers (e.g. wallets) are given the built transaction group and may change what they sign, such as in Pera PQ account scenarios. When they do, the transactions that reach the network are not the ones that were built (different txn and group IDs)

The group was sent successfully, but everything reported afterwards was derived from the pre-mutation transactions. The `pendingTransactionInformation` lookups behind `confirmations` then failed with:

```
404 could not find the transaction in the transaction pool or in the last 1000 confirmed rounds
```

for a transaction ID that was never on chain, and `txIds`, `transactions` and `groupId` were stale in the same way.

This currently manifests on Lora if a txn wizard transaction is mutated for a PQ account.

## Fix

New internal helper `resolveSignedTransactions` decodes the signed transactions and returns those, keeping the built transaction object whenever its ID is unchanged. It is not re-exported from `src/transaction/index.ts`, so the public API is unchanged.

It is applied at three sites:

- `sendAtomicTransactionComposer` — `txIds`, `transactions`, `groupId`, `confirmations` and the "Sent transaction ID" / "Group transaction" logs, plus `err.sentTransactions` on the error path (which `AppClient.handleCallErrors` matches transaction IDs against).
- `TransactionComposer.simulate()` — same stale `transactions` / `txIds` / `groupId`.
- The legacy `sendTransaction` — logged and waited on the pre-signing transaction ID.

In the composer paths the signatures were already gathered by `execute()` / `simulate()`, so `atc.gatherSignatures()` returns them from the composer's cache; signers are never prompted twice.

`app-deployer.ts` and `app-client.ts` consume `result.transactions` and `err.sentTransactions`, so they are correct as a consequence.

## Tests

Three tests in `src/types/composer.spec.ts` use a signer that adds a note to what it is given (single transaction, a regrouped group of two, and `simulate()`). Without the fix all three fail, the two send cases with the 404 above.

## Known limitation, not addressed here

`TransactionComposer.send()` computes `waitRounds` from the built group's `lastValid` before signing, so a signer that changes `lastValid` still gets the wrong wait window.  Fixing it means signing before that calculation, which would move the wallet prompt ahead of the `Config.debug && Config.traceAll` simulate. Not sure if we want to change that right now, as there is no known usage scenario that changes lastValid currently, or if we'd rather open an issue about this. @joe-p


joe-p (approved):

joe-p:
@tasosbit can you sign your commit?

tasosbit:
@joe-p I did but I clobbered your audit fix :cold_face: can you push it again? 

edit: switched my push force alias to use --force-with-lease, should not happen again

